PR #5, opened 2024-10-11
Rename paas_app_charmer -> paas_charm

javierdelapuente:
Applicable spec: <link>

### Rationale

The goal of this PR is to rename `paas-app-charmer` to `paas-charm` (both the repo and the modules). The original repository is [paas-app-charmer repo](https://github.com/canonical/paas-app-charmer).

To be backward compatible, it is necessary to keep the following classes (see the templates in [charmcraft](https://github.com/canonical/charmcraft/tree/main/charmcraft/templates)):
-  `paas_app_charmer.django.Charm`
-  `paas_app_charmer.fastapi.Charm`
-  `paas_app_charmer.flask.Charm`
-  `paas_app_charmer.go.Charm`

The project structure has been changed from `flat layout` to `src layout` as it there are now two top-level packages. 

As this is a new project, the version will start with `1.0.0`.

### Checklist


amandahla (dismissed):

pyproject.toml
```diff
@@ -1,9 +1,9 @@
 # Copyright 2024 Canonical Ltd.
 # See LICENSE file for licensing details.
 [project]
-name = "paas-app-charmer"
-version = "1.3.1"
-description = "Companion library for PaaS app charmer."
+name = "paas-charm"
+version = "1.0.0"
```
jdkandersson, line 5: I think we should keep the version the same, I believe charmcraft has a minimum version specified in the template and we should probably keep updating that to make sure people get the latest when they start

src/paas_app_charmer/fastapi/charm.py
```diff
@@ -0,0 +1,11 @@
...
+# See LICENSE file for licensing details.
+
+
+"""Deprecated FastAPI Charm service."""
+
+import paas_charm.fastapi.charm
+
+
+class Charm(paas_charm.fastapi.charm.Charm):
+    """Deprecated FastAPI charm."""
```
jdkandersson, line 11: Perhaps we include a note here that these will be maintained for the life of 24.04 and 22.04? For 26.04 they would be removed

jdkandersson: For the other deprecated classes as well

src/paas_app_charmer/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+# See LICENSE file for licensing details.
+
+"""Deprecated entrypoints for charms.
+
+This module can be removed when paas_charm>=2.0
+"""
+
+import warnings
+
+warnings.warn("paas_app_charmer package is deprecated", DeprecationWarning, stacklevel=2)
```
jdkandersson, line 11: Perhaps include a note here on the maintenance window which is that it will be maintained for the lifetime of 22.04 and 24.04

jdkandersson: And removed in 26.04

tests/unit/test_deprecated.py
jdkandersson, line 1: Not sure if it is there already, could we have an integration test for these as well? I.e., a charm with `paas-app-charmer` in requirements.txt and check that everything works fine

pyproject.toml
javierdelapuente: I initially thought that way, however it is a totally new project and looked strange in Pypi.

charmcraft will have the dependency we prefer (initially I have put `paas-charm>=1.0,<2`), see https://github.com/canonical/charmcraft/pull/1961 for a possibility.

As the old paas-app-charmer depends on `paas-charm<2`, the only real requirement for the initial release is that the version is in the major 1.

src/paas_app_charmer/fastapi/charm.py
javierdelapuente: Sure. Let's be explicit about that.

The real dependency is with the major version 1. Old projects (from 22.04 and 24.04) that do `charmcraft pack` will get the latest 1.X.X version of this project, so the way of breaking the dependency is changing the major version to 2 for 26.04 bases (and state it that way in charmcraft). 

If the major is not updated when bases 26.04 are in use, I think we will have to maintain the deprecated classes (and the backward compatibility). When 26.04 is added, if we release version 2, we will have to maintain two different branches.

src/paas_app_charmer/__init__.py
javierdelapuente: I added in the module docstring the comment about 22.04 and 24.04.

In the case of the warning for user, I am not sure about the message. I mean, when we release v2 of the library, all v1 may be changed/deprecated so not sure that the extra info will be useful.

tests/unit/test_deprecated.py
javierdelapuente: I tried to avoid the tests, because it is a heavy test and you cannot really test `paas-app-charmer`, as you would be testing the Pypi package of `paas-charm`. 

I think if we want that test, it should be in `charmcraft`.

jdkandersson: let's make use of test.pypi.org in another story and run an integration test there

javierdelapuente: Added ISD-2515 for this.

jdkandersson (approved):

amandahla (approved):

PR #9, opened 2024-10-17
Add base_url to list of django allowed hosts

javierdelapuente:
Applicable spec: <link>

### Overview

Adding the ingress url or the k8s service url to the list of allowed hosts by default.

This will simplify the usage of the charm, as the ingress of the hostname will be most times needed as an allowed hosts. If it is not, the user just can add another element to the list of allowed hosts.

### Checklist

### DIFF
--- a/src/paas_charm/django/charm.py
+++ b/src/paas_charm/django/charm.py
@@ -80,9 +80,13 @@ class Charm(GunicornBase):
 
         The method is overridden to inject the base_url, that can be an ingress URL or a k8s svc
         url, to the list of allowed hosts.
+
+        Returns:
+             Framework related configurations.
         """
         base_model = super().get_framework_config()
-        base_model.allowed_hosts.append(self._base_url)
+        # base_model can be downcasted to a DjangoConfig, and allowed_hosts is really a list.
+        base_model.allowed_hosts.append(self._base_url)  # type: ignore
         return base_model
 
     def is_ready(self) -> bool:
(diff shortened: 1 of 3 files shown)


src/paas_charm/django/charm.py
```diff
@@ -75,6 +75,20 @@ def get_cos_dir(self) -> str:
...
+
+        The method is overridden to inject the base_url, that can be an ingress URL or a k8s svc
+        url, to the list of allowed hosts.
+
+        Returns:
+             Framework related configurations.
+        """
+        base_model = super().get_framework_config()
+        # base_model can be downcasted to a DjangoConfig, and allowed_hosts is really a list.
+        base_model.allowed_hosts.append(self._base_url)  # type: ignore
```
amandahla, line 89: I'm a bit out of the loop, so this might be a random question: what if the ingress URL changes?

jdkandersson: The point is good, we need to make sure we reload the app if the ingress changes and make sure that we reload the allowed hosts

javierdelapuente: As `paas-charm` is designed in a holistic way, every event that changes something will call a replan, and as the env vars change, the service will be restarted. As there is no harm in an extra test I have added a ~~"unit"~~ test for it.

(Besides I have fixed a bug that I did not see because of the '*', I updated one integ test so it fails if the allowed_host is not correct).

jdkandersson (approved):

amandahla (approved):

PR #11, opened 2024-11-07
Feat(): Enable async workers for flask and django

alithethird:
Applicable spec: <link>

### Checklist


requirements.txt
```diff
@@ -2,3 +2,4 @@ cosl
 jsonschema >=4.19,<4.20
 ops >= 2.6
 pydantic==2.9.2
+websockets < 14.0
```
cbartz, line 5: Isn't this only a requirement for the integration tests (python-libjuju)? If yes, it should better be placed in tox.ini

src/paas_charm/_gunicorn/charm.py
```diff
@@ -23,6 +26,67 @@ def _workload_config(self) -> WorkloadConfig:
             framework_name=self._framework_name, unit_name=self.unit.name
         )
 
+    def check_gevent_package(self) -> bool:
+        """Check that gevent is installed.
+
+        Returns:
+            True if gevent is installed.
+        """
+        ddddd = self._container.exec(["pip", "list"])
```
cbartz, line 35: May I know the reason for this variable naming?

src/paas_charm/_gunicorn/charm.py
```diff
@@ -23,6 +26,67 @@ def _workload_config(self) -> WorkloadConfig:
...
+    def create_webserver_config(self) -> WebserverConfig:
+        """Create a WebserverConfig instance from the charm config.
+
+        Returns:
+            A new WebserverConfig instance.
+
+        Raises:
+            CharmConfigInvalidError: if the charm configuration is not valid.
+        """
+        _webserver_config: WebserverConfig = WebserverConfig.from_charm_config(dict(self.config))
```
cbartz, line 48: Variable with underscore in a method is unusual . What is the reason?

cbartz (commented):
Partial review from mobile phone

src/paas_charm/_gunicorn/charm.py
alithethird: I was testing first so did not think about a variable name but it turns out I forgot to name it properly before committing. Thank you for noticing!

src/paas_charm/_gunicorn/charm.py
alithethird: No reason. Just a complete error.

src/paas_charm/_gunicorn/charm.py
```diff
@@ -23,6 +26,67 @@ def _workload_config(self) -> WorkloadConfig:
...
+                    self._framework_name,
+                    self._framework_name,
+                )
+                raise CharmConfigInvalidError(
+                    "Only 'gevent' and 'sync' are allowed. "
+                    "https://documentation.ubuntu.com/rockcraft/en/latest"
+                    f"/reference/extensions/{self._framework_name}-"
+                    f"framework/#parts-{self._framework_name}-"
+                    "framework-async-dependencies"
+                )
```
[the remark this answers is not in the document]
alithethird: I will research that. Thank you!

examples/django/django_async_app/rockcraft.yaml
jdkandersson, line 1: Perhaps I missed it, I didn’t see any async def functions for views?

examples/flask/test_async_rock/app.py
jdkandersson, line 1: Could this be simplified and perhaps just include a path handled by an async def function?

src/paas_charm/_gunicorn/charm.py
```diff
@@ -23,6 +25,58 @@ def _workload_config(self) -> WorkloadConfig:
...
 
+    def check_gevent_package(self) -> bool:
+        """Check that gevent is installed.
+
+        Returns:
+            True if gevent is installed.
+        """
+        pip_list_command = self._container.exec(["pip", "list"])
+        list_output = pip_list_command.wait_output()[0]
+        return "gevent" in list_output
```
jdkandersson, line 36: I think this is a bit brittle as, for example, a package could be called gevents and this would pass but gevent is not installed

tests/integration/django/test_workers.py
```diff
@@ -0,0 +1,72 @@
...
+import requests
+from juju.application import Application
+from juju.model import Model
+from juju.utils import block_until
+from pytest_operator.plugin import OpsTest
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize(
```
jdkandersson, line 21: Do you think these could be more unit tests of the config validation?

jdkandersson: Also, should we send a request to an app and make sure we get a response?

src/paas_charm/_gunicorn/charm.py
alithethird: Thank you for pointing this out. Fixed it. Now I check it by importing the package and calling `__version__` on it.

tests/integration/django/test_workers.py
alithethird: Moved the config tests to unit and add a integration test that sends a request to the app

examples/django/django_async_app/rockcraft.yaml
alithethird: Async def with gunicorn async workers are not recommended. We researched it with Greg and decided the functions should be sync when using gevent.

examples/django/django_async_app/django_async_app/django_async_app/settings.py
```diff
@@ -0,0 +1,130 @@
...
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/5.0/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = os.environ.get("DJANGO_SECRET_KEY", "secret")
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = os.environ.get("DJANGO_DEBUG", "true") == "true"
+
+ALLOWED_HOSTS = json.loads(os.environ.get("DJANGO_ALLOWED_HOSTS", '["*"]'))
```
javierdelapuente, line 34: there will be always a DJANGO_ALLOWED_HOSTS, and it should contain at least the hostname, so this could be changed to `... json.loads(os.environ["DJANGO_ALLOWED_HOSTS"]`)`

examples/django/django_async_app/django_async_app/migrate.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/bash
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+python3 manage.py migrate
```
javierdelapuente, line 5: We do not need this file. It will work without it.

src/paas_charm/_gunicorn/charm.py
```diff
@@ -23,6 +25,60 @@ def _workload_config(self) -> WorkloadConfig:
             framework_name=self._framework_name, unit_name=self.unit.name
         )
 
+    def check_gevent_package(self) -> bool:
```
javierdelapuente, line 28: This function and the following one could be private (they are not called anywhere else and are specific for Django). Besides, I would put it in the end of the file (see https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#function-and-method-Ordering)

src/paas_charm/_gunicorn/charm.py
```diff
@@ -23,6 +25,60 @@ def _workload_config(self) -> WorkloadConfig:
...
+        """Check that gevent is installed.
+
+        Returns:
+            True if gevent is installed.
+        """
+        pip_list_command = self._container.exec(
+            ["python3", "-c", "'import gevent;print(gevent.__version__)'"]
+        )
+        list_output = pip_list_command.wait_output()[0]
```
javierdelapuente, line 37: it the python3 command fails, it will exit with an error code that is not zero, and this will raise. I think the unit tests also do not test that the return code is not 0.

javierdelapuente (commented):
nice! a few comments

weiiwang01 (approved):

jdkandersson (approved):

tests/unit/flask/constants.py
```diff
@@ -8,7 +8,7 @@
         "flask": {
             "override": "replace",
             "startup": "enabled",
-            "command": f"/bin/python3 -m gunicorn -c /flask/gunicorn.conf.py app:app",
+            "command": f"/bin/python3 -m gunicorn -c /flask/gunicorn.conf.py app:app -k sync",
```
jdkandersson, line 11: Are there any tests without the `-k sync` for backwards compatibility please?

alithethird: The latest paas-charm instance automatically adds `-k sync` to command as can be seen [here](https://github.com/canonical/paas-charm/blob/88d898f105c40294c56e9b8d198bf4a9d4df108d/src/paas_charm/_gunicorn/wsgi_app.py#L51). How can I test it for backwards compatibility with this?
Also  thanks to you I just realised I need to put an empty space before adding the `-k sync` for backwards compatibility. Thanks!

jdkandersson: Is this what is expected to be added?

PR #20, opened 2024-12-16
Tempo tracing

alithethird:
### Overview

Adds new integration Tempo to the charm. Creates a class called `TempoParameters` that handles the parameters. There are only 2 parameters:
- Endpoint
- Service name

Endpoint comes from the integration and service name is generated by reading the charm name from the `charmcraft.yaml` file.

### Checklist


src/paas_charm/charm.py
```diff
@@ -117,6 +127,14 @@ def __init__(self, framework: ops.Framework, framework_name: str) -> None:
         else:
             self._rabbitmq = None
 
+        if "tracing" in requires and requires["tracing"].interface_name == "tracing":
+            self._tracing = TracingEndpointRequirer(
+                self, relation_name="tracing", protocols=["otlp_http"]
+            )
+            # add self.framework.observe for relation changed and departed
```
jdkandersson, line 134: For the interface spec, these should be enumerated out

src/paas_charm/charm.py
```diff
@@ -349,6 +372,12 @@ def _create_charm_state(self) -> CharmState:
                 for k, v in charm_config.items()
             },
         )
+        tracing_relation_data = None
+        if self._tracing and self._tracing.is_ready():
+            tracing_relation_data = TempoParameters(
+                endpoint=self._tracing.get_endpoint(protocol="otlp_http"),
+                service_name=f"{self.framework.meta.name}-charm",
```
jdkandersson, line 379: Let's call the service name the `<application name>-app`

src/paas_charm/charm_state.py
```diff
@@ -89,6 +89,7 @@ def from_charm(  # pylint: disable=too-many-arguments
         s3_connection_info: dict[str, str] | None = None,
         saml_relation_data: typing.MutableMapping[str, str] | None = None,
         rabbitmq_uri: str | None = None,
+        tracing_relation_data: "TempoParameters | None" = None,
```
jdkandersson, line 92: I think the string is needed here

src/paas_charm/charm_state.py
```diff
@@ -209,13 +213,16 @@ class IntegrationsState:
         s3_parameters: S3 parameters.
         saml_parameters: SAML parameters.
         rabbitmq_uri: RabbitMQ uri.
+        tracing_relation_data: The tracing uri provided by the Tempo coordinator charm
+        and charm name.
```
jdkandersson, line 217: This should be indented 4 more spaces

src/paas_charm/charm_state.py
```diff
@@ -238,6 +246,8 @@ def build(  # pylint: disable=too-many-arguments
             s3_connection_info: S3 connection info from S3 lib.
             saml_relation_data: Saml relation data from saml lib.
             rabbitmq_uri: RabbitMQ uri.
+            tracing_relation_data: The tracing uri provided by the Tempo coordinator charm
+        and charm name.
```
jdkandersson, line 250: Indentation

src/paas_charm/charm_state.py
```diff
@@ -209,13 +213,16 @@ class IntegrationsState:
...
+        tracing_relation_data: The tracing uri provided by the Tempo coordinator charm
+        and charm name.
     """
 
     redis_uri: str | None = None
     databases_uris: dict[str, str] = field(default_factory=dict)
     s3_parameters: "S3Parameters | None" = None
     saml_parameters: "SamlParameters | None" = None
     rabbitmq_uri: str | None = None
+    tracing_relation_data: "TempoParameters | None" = None
```
jdkandersson, line 225: Perhaps we need to move the other class definitions up

alithethird: Moved the classes to the top of file and removed `"`.

src/paas_charm/charm_state.py
alithethird: Moved the classes to the top of file and removed ".

src/paas_charm/charm.py
```diff
@@ -266,6 +284,8 @@ def is_ready(self) -> bool:
 
         return True
 
+    # pylint: disable=too-many-branches
+    # disabled because we have too many possible integrations for the workload.
     # Pending to refactor all integrations
     def _missing_required_integrations(self, charm_state: CharmState) -> list[str]:  # noqa: C901
```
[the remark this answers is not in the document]
alithethird: Created the Generator functions and yielded them in `_missing_required_integrations` function.

src/paas_charm/charm.py
```diff
@@ -70,6 +79,8 @@ def _create_app(self) -> App:
 
     on = RedisRelationCharmEvents()
 
+    # pylint: disable=too-many-statements
+    # disabled because we have too many possible integrations for the workload.
```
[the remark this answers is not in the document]
alithethird: Created init functions for each integration.

src/paas_charm/app.py
```diff
@@ -259,6 +259,15 @@ def map_integrations_to_env(integrations: IntegrationsState, prefix: str = "") -
...
         env.update(interface_envvars)
+    if integrations.tracing_relation_data:
+        env.update(
+            (k, v)
+            for k, v in (
+                ("OTEL_SERVICE_NAME", integrations.tracing_relation_data.service_name),
+                ("OTEL_EXPORTER_OTLP_ENDPOINT", integrations.tracing_relation_data.endpoint),
+            )
+            if v is not None
+        )
```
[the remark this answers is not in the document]
alithethird: Simplified.

tests/integration/integrations/test_tracing.py
```diff
@@ -0,0 +1,83 @@
...
+
+@pytest.mark.parametrize(
+    "tracing_app, port",
+    [
+        ("flask_tracing_app", 8000),
+        ("django_tracing_app", 8000),
+        ("fastapi_tracing_app", 8080),
+        ("go_tracing_app", 8080),
+    ],
+)
```
gregory-schiano, line 30: I think you can use `indirect` fixture by using `indirect=["tracing_app"]`

This will avoid calling getfixturevalue

tests/integration/integrations/test_tracing.py
```diff
@@ -0,0 +1,83 @@
...
+
+    try:
+        tempo_app = await request.getfixturevalue("tempo_app")
+    except Exception as e:
+        logger.info(f"Tempo is already deployed  {e}")
+
+    tracing_app = request.getfixturevalue(tracing_app)
+    idle_list = [tracing_app.name]
+
+    if tracing_app != "flask_tracing_app":
```
gregory-schiano, line 54: Should probably be `tracing_app.name`

tests/unit/flask/test_tracing.py
```diff
@@ -0,0 +1,38 @@
...
+import unittest.mock
+
+import ops
+import pytest
+from ops.testing import Harness
+
+from .constants import DEFAULT_LAYER, FLASK_CONTAINER_NAME
+
+
+def test_tracing_relation(harness: Harness):
```
gregory-schiano, line 15: I'd add a test with tracing disabled and ensure the envvars aren't injected
And I think you can also add a test in test_webserver.py that ensure the generated gunicorn config contains the tracing methods

src/paas_charm/_gunicorn/webserver.py
```diff
@@ -171,6 +171,32 @@ def _config(self) -> str:
...
+        if services:
+            service_framework = services.get(self._workload_config.framework, None)
+            if service_framework:
+                framework_environments = service_framework.get("environment", None)
+
+        if framework_environments and framework_environments.get(
+            "OTEL_EXPORTER_OTLP_ENDPOINT", None
+        ):
+            config += textwrap.dedent(
+                """\
```
javierdelapuente, line 186: Maybe it could be simpler to put the configuration in a jinja template or similar, as it it slowly getting complex

src/paas_charm/charm_state.py
```diff
@@ -20,6 +20,95 @@
 logger = logging.getLogger(__name__)
 
 
+class TempoParameters(BaseModel):
```
javierdelapuente, line 23: Although ordering the classes so everything is defined before is used has advantages, it also has disadvantages.

The disadvantage of putting this class as the first one of the file, is that when opening it, it takes a while to realise that the most important class is CharmState and what it contains. In this way readibility is decreased. This is similar to https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#function-and-method-Ordering

tests/integration/integrations/conftest.py
```diff
@@ -0,0 +1,267 @@
...
+from minio import Minio
+from ops import JujuVersion
+from pytest import Config, FixtureRequest
+from pytest_operator.plugin import OpsTest
+
+from tests.integration.helpers import inject_charm_config, inject_venv
+
+PROJECT_ROOT = pathlib.Path(__file__).parent.parent.parent.parent
+
+import nest_asyncio
```
javierdelapuente, line 25: is this needed?

tests/integration/integrations/test_tracing.py
```diff
@@ -0,0 +1,83 @@
...
+    get_unit_ips,
+):
+    """
+    arrange: Deploy Tempo cluster, app to test and postgres if required.
+    act: Send 5 requests to the app.
+    assert: Tempo should have tracing info about the app.
+    """
+
+    try:
+        tempo_app = await request.getfixturevalue("tempo_app")
```
javierdelapuente, line 48: why not having a fixture for this?

tests/integration/integrations/test_tracing.py
```diff
@@ -0,0 +1,83 @@
...
+    except Exception as e:
+        logger.info(f"Tempo is already deployed  {e}")
+
+    idle_list = [tracing_app.name]
+
+    if tracing_app.name != "flask-tracing-k8s":
+        try:
+            postgresql_app = request.getfixturevalue("postgresql_k8s")
```
javierdelapuente, line 56: could postgresq_k8s be integrated in the django_tracing_app /fastapi_tracing_app instead?

tests/unit/flask/test_tracing.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Flask charm tracing relation unit tests."""
+
+import unittest.mock
+
+import ops
+import pytest
+from ops.testing import Harness
```
javierdelapuente, line 10: would it be very difficult to write the new tests using scenario as harness is `legacy`?

javierdelapuente (commented):
Good work!

Just a question (maybe I it is already discussed). As the examples are similar to the base ones, why not updating the base examples instead of adding new ones?

tests/integration/integrations/test_tracing.py
alithethird: I tried to use this but it turns out it doesn't call the `flask_tracing_app` when you use `indirect`. What it does is calls the `tracing_app` fixture with a parameter that is `flask_tracing_app`. If I do that I have to create a fixture like the following:

```python
@pytest.fixture(name="tracing_app")
async def tracing_app_fixture(request):
    return await request.getfixturevalue(request.param)
```

I think this is worse then just using the `getfixturevalue` inside the test. Since in both cases we use that function and in `indirect` method we add more lines.

alithethird: But if you think `indirect` method is better I can implement it.

tests/integration/integrations/test_tracing.py
alithethird: fixed re-assigning the variable issues. Thanks for pointing out

tests/unit/flask/test_tracing.py
alithethird: done.

alithethird: I will also try to add a test for the following lines:
```python
try:
    # pylint: disable=ungrouped-imports
    from charms.tempo_coordinator_k8s.v0.tracing import TracingEndpointRequirer
except ImportError:
    logger.exception(
        "Missing charm library, please run "
        "`charmcraft fetch-lib charms.tempo_coordinator_k8s.v0.tracing`"
    )

```

tests/integration/integrations/test_tracing.py
alithethird: This would copy the same piece of code to 3 different places. WDYT @jdkandersson ?

src/paas_charm/charm_state.py
alithethird: WDYT? @jdkandersson

tests/integration/integrations/test_tracing.py
javierdelapuente: There will be two more lines in each those three fixtures. However in this `if` there are already 8 lines. Besides, all the fixtures will be complete and equal, and also could be reused in other tests (currently, for example the `flask-tracing-k8s` waits for the app to be active but not the others).

src/paas_charm/charm_state.py
jdkandersson: The reason is around that these are classes that other classes depend on so you have to define them first I think

jdkandersson:
I think I agree with Javi that we should just use existing examples and add tempo to them. Adding a lot of extra examples is a lot to maintain

src/paas_charm/charm_state.py
javierdelapuente: In our case the difference is this:
```
   - s3_parameters: "S3Parameters | None" = None
   - saml_parameters: "SamlParameters | None" = None
   + s3_parameters: S3Parameters | None = None
   + saml_parameters: SamlParameters | None = None
```
the type hints have to be defined with strings. It is of course a matter of preference. My opinion leans toward using strings (that are checked by the linter), although they are not as elegant.

tests/unit/flask/test_tracing.py
alithethird: For this PR, we will skip this. In the next PR(Kafka integration) I will check out scenario. If we can gradually move into scenario I will do that. If it requires all the harness tests to be moved into that we might need to wait a little and create all in 1 PR.

src/paas_charm/_gunicorn/webserver.py
alithethird: done.

tests/integration/integrations/test_tracing.py
alithethird: When I use this as a fixture it tries to deploy before the running the marker that checks Juju version and in Juju 3.3 test it fails to deploy.

tests/unit/flask/test_tracing.py
alithethird: Researched and no one tested for this yet. We will do this test in a future PR. We also need to test other libraries too. Created a [story](https://warthogs.atlassian.net/browse/ISD-2918) for it.

tests/integration/integrations/conftest.py
alithethird: yeap

tests/integration/integrations/test_tracing.py
alithethird: done

tests/unit/flask/test_tracing.py
javierdelapuente: Thanks! just to not forget that harness is `legacy` and not recommended anymore so we can gradually migrate.

tests/integration/integrations/test_tracing.py
javierdelapuente: sorry, I did not understand that. If the fixture is not requested, I think it should not be run, isn't it?

alithethird: When I use it as a regular fixture like the following:
```python
@pytest.mark.skip_juju_version("3.4")
test_tracing(tempo_app, ...):
  ...
```

it runs the `tempo_app` fixture before checking the juju version which results in error when juju version is lower then 3.4

javierdelapuente: Interesting. Yeah, kind of makes sense, as the temp_app is defined as "session" scoped, it probably runs before the skip juju one...

src/paas_charm/templates/gunicorn.conf.py.j2
jdkandersson, line 1:
```python
{% if enable_tracing %}
from opentelemetry import trace
from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
    OTLPSpanExporter,
)
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import BatchSpanProcessor

{% endif %}
bind = ['0.0.0.0:{{ workload_port }}']
chdir = '{{ workload_app_dir }}'
accesslog = '{{ access_log }}'
errorlog = '{{ error_log }}'
statsd_host = '{{ statsd_host }}'
{%- for key, value in config_entries.items() -%}
{% endfor %}
{% if enable_tracing %}

def post_fork(server, worker):
    trace.set_tracer_provider(TracerProvider())
    span_processor = BatchSpanProcessor(OTLPSpanExporter())
    trace.get_tracer_provider().add_span_processor(span_processor)
{% endif %}
```

At the end, look at a generated file and check that black wouldn't make changes

src/paas_charm/_gunicorn/webserver.py
```diff
@@ -146,32 +146,47 @@ def _config(self) -> str:
...
+        plan = self._container.get_plan().to_dict()
+        services = plan.get("services", None)
+        if services:
+            service_framework = services.get(self._workload_config.framework, None)
+            if service_framework:
+                framework_environments = service_framework.get("environment", None)
+        if framework_environments and framework_environments.get(
+            "OTEL_EXPORTER_OTLP_ENDPOINT", None
+        ):
+            enable_tracing = True
```
jdkandersson, line 176: I would suggest that we introduce a new attribute either in webserver or workload config to indicate whether tracing is enabled

src/paas_charm/_gunicorn/webserver.py
jdkandersson, line 1: The behaviour of whether tracing is enabled should only be checking whether the integration is present. If the integration is present and the library is missing the charm should go into blocked state and tell the user to install the library/ repack the charm

src/paas_charm/charm.py
```diff
@@ -164,6 +146,124 @@ def __init__(self, framework: ops.Framework, framework_name: str) -> None:
...
+            try:
+                _redis = RedisRequires(charm=self, relation_name="redis")
+                self.framework.observe(
+                    self.on.redis_relation_updated, self._on_redis_relation_updated
+                )
+            except NameError:
+                logger.exception(
+                    "Missing charm library,                               "
+                    "please run `charmcraft fetch-lib charms.redis_k8s.v0.redis`"
+                )
```
jdkandersson, line 169: We'll need something like this for all the integrations where the libraries might be missing. Check the formatting of the log message

We also need to introduce a check around https://github.com/canonical/paas-charm/blob/38d7f1e2ea36f252a0a1874cd928c092310435c3/src/paas_charm/charm.py#L359 where the logic should be if the integration is detected but the library is missing, the charm goes into blocked state telling the user to include the library in the charm

jdkandersson: This is needed for:

* S3
* SAML
* Tracing

src/paas_charm/charm.py
```diff
@@ -349,6 +482,12 @@ def _create_charm_state(self) -> CharmState:
                 for k, v in charm_config.items()
             },
         )
+        tracing_relation_data = None
+        if self._tracing and self._tracing.is_ready():
+            tracing_relation_data = TempoParameters(
+                endpoint=f'{self._tracing.get_endpoint(protocol="otlp_http")}',
+                service_name=self.app.name,
+            )
```
jdkandersson, line 490: Instead of this, I would pass the tracing object and the app name to `from_charm` rather than this object

src/paas_charm/charm_state.py
```diff
@@ -209,13 +302,16 @@ class IntegrationsState:
         s3_parameters: S3 parameters.
         saml_parameters: SAML parameters.
         rabbitmq_uri: RabbitMQ uri.
+        tracing_relation_data: The tracing uri provided by the Tempo coordinator charm
+            and charm name.
```
jdkandersson, line 306: let's rename this to tempo parameters

src/paas_charm/_gunicorn/webserver.py
alithethird: I added a new attribute to the workload config. It seemed more suitable then webserver config since the Tempo relation is not about the webserver.

examples/fastapi/charm/lib/charms/traefik_k8s/v2/ingress.py
```diff
@@ -73,7 +72,7 @@ def _on_ingress_revoked(self, event: IngressPerAppRevokedEvent):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 14
+LIBPATCH = 13
```
jdkandersson, line 75: it looks like we're going backwards on the version here? Is version 14 available, or perhaps an even later version?

examples/flask/lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -331,14 +331,10 @@ def _on_topic_requested(self, event: TopicRequestedEvent):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 40
+LIBPATCH = 36
```
jdkandersson, line 334: Also looks like the version is being downgraded here

examples/fastapi/charm/lib/charms/traefik_k8s/v2/ingress.py
gregory-schiano: Since the library is in an example folder, they're not automatically updated, all the other examples are using version 13, so Ali aligned.
I think it should be good to:
- Upgrade the libraries manually once to get the latest
- Add a story to find a way to automate this library upgrade via juju action or renovate

examples/flask/lib/charms/data_platform_libs/v0/data_interfaces.py
gregory-schiano: Closing the comment, answer in the ingress library comment

requirements.txt
gregory-schiano, line 1: We should pin dependencies in `requirements.txt`

gregory-schiano: Could be tackled in another PR as it's not related to Tempo tracing

src/paas_charm/charm_state.py
gregory-schiano: We're even not consistent because `IntegrationState` is declared after `CharmState` but used in a hint in `CharmState`
I don't like when a linting/hinting tool dictate the structure of the code (in the sense that we do things to please it). The fact that the class must be declared before to be used without strings in the typehints is a limitation of mypy and should force us to go to a less natural ordering.
Usually I'm in favor of ordering in the natural code execution order, as when you reason about the code, you're following it in the order of execution. And that's the order defined in our coding guideline. Generally I'm pretty flexible there, since a small data object (like HTTPProxy) before the CharmState doesn't alter too much the readability, and allows to immediately know Proxy is properly manage in the charm. But when the CharmState becomes more complex with embedded state object, I prefer reading the charmstate first and drill down.

So as a conclusion:
- Simple charmstate with few data object, I'm fine having them defined first
- More complex charmstate, I'd rather have the charmstate first

src/paas_charm/charm.py
```diff
@@ -498,7 +492,7 @@ def _create_charm_state(self) -> CharmState:
             s3_connection_info=self._s3.get_s3_connection_info() if self._s3 else None,
             saml_relation_data=saml_relation_data,
             rabbitmq_uri=self._rabbitmq.rabbitmq_uri() if self._rabbitmq else None,
-            tracing_relation_data=tracing_relation_data,
+            tempo_parameters=TempoParameters.from_charm(name=self.app.name, tracing=self._tracing),
```
gregory-schiano, line 495: I feel like the approach is incoherent compared to how other optional relations are treated. S3 and SAML for example don't use a `from_charm` passed to the charmstate.
Instead they're passing their parameter to the `CharmState` and then it leverages the `IntegrationState.build` to generate the `Parameters` objects.

So what I suggest:
- Check if the relation exists is done in this method for other relation
- Similarly to `saml_relation_data`, build a dict that is either `None` or contains the data required for the TempoParameters
- Pass that dict to the `CharmState.from_charm`, that will use the `IntegrationState.build` with these parameters, to return the `TempoParameters`

examples/fastapi/charm/lib/charms/traefik_k8s/v2/ingress.py
gregory-schiano: Could done in a separate PR as it's unrelated to this PR

src/paas_charm/charm_state.py
```diff
@@ -338,10 +205,85 @@ class IntegrationsState:
...
+            CharmConfigInvalidError: If some parameter in invalid.
+        """
+        if saml_relation_data is not None:
+            try:
+                relation_parameter = parameter_type(**saml_relation_data)
+            except ValidationError as exc:
+                error_message = build_validation_error_message(exc)
+                raise CharmConfigInvalidError(f"Invalid configuration: {error_message}") from exc
+        else:
+            relation_parameter = None
```
gregory-schiano, line 238: You can initialize relation_parameter before the if statement to None which will avoid the else

gregory-schiano: Or be defensive
```python
if saml_relation_data is None:
   return None
```

src/paas_charm/charm_state.py
```diff
@@ -338,10 +205,85 @@ class IntegrationsState:
...
     databases_uris: dict[str, str] = field(default_factory=dict)
-    s3_parameters: S3Parameters | None = None
-    saml_parameters: SamlParameters | None = None
+    s3_parameters: "S3Parameters | None" = None
+    saml_parameters: "SamlParameters | None" = None
     rabbitmq_uri: str | None = None
-    tempo_parameters: TempoParameters | None = None
+    tempo_parameters: "TempoParameters | None" = None
+
+    @classmethod
```
gregory-schiano, line 213: This method and the next 2 could be module methods

src/paas_charm/charm_state.py
```diff
@@ -338,10 +205,85 @@ class IntegrationsState:
...
+
+        Raises:
+            CharmConfigInvalidError: If some parameter in invalid.
+        """
+        if relation_data:
+            try:
+                relation_parameter = parameter_type(**relation_data)
+            except ValidationError as exc:
+                error_message = build_validation_error_message(exc)
+                raise CharmConfigInvalidError(f"Invalid configuration: {error_message}") from exc
```
gregory-schiano, line 264: Since the error message becomes generic we should make sure that the error message now doesn't degrade user experience

src/paas_charm/charm_state.py
```diff
@@ -238,44 +232,22 @@ def build(  # pylint: disable=too-many-arguments
...
-            try:
-                saml_parameters = SamlParameters(**saml_relation_data)
-            except ValidationError as exc:
-                error_message = build_validation_error_message(exc)
-                raise CharmConfigInvalidError(
-                    f"Invalid Saml configuration: {error_message}"
-                ) from exc
-        else:
-            saml_parameters = None
+        s3_parameters, saml_parameters, tempo_parameters = list(
```
javierdelapuente, line 241: I think it is a good idea to try to improve this code. However, it maybe be an advantage to do and discuss it in another (small) PR.

requirements.txt
```diff
@@ -2,4 +2,4 @@ cosl
 jsonschema >=4.23,<4.24
 ops >= 2.6
 pydantic==2.10.6
-Jinja2
+Jinja2==3.1.5
```
jdkandersson, line 5: Final new line please

src/paas_charm/charm.py
```diff
@@ -492,7 +502,7 @@ def _create_charm_state(self) -> CharmState:
             s3_connection_info=self._s3.get_s3_connection_info() if self._s3 else None,
             saml_relation_data=saml_relation_data,
             rabbitmq_uri=self._rabbitmq.rabbitmq_uri() if self._rabbitmq else None,
-            tempo_parameters=TempoParameters.from_charm(name=self.app.name, tracing=self._tracing),
+            tempo_relation_data=tempo_relation_data,
```
jdkandersson, line 505: Let's change this to pass the tracing requires object and the app name and figure out the rest in charm state

jdkandersson (approved):

javierdelapuente (approved):

PR #32, opened 2025-01-30
Smtp integration

alithethird:
### Overview

Add SMTP integration to 12 Factor apps.

### Rationale

User story:
  As a user, I need to email my clients for application notifications. I would need SMTP credentials to add email integration to my app so that I can send emails to my clients using the credentials. 

### Juju Events Changes

Observe `smtp_data_available` event.

### Module Changes

Add 3 new classes to store SMTP credentials. 
- SmtpParameters
- AuthType
- TransportSecurity

These classes are copied from smtp library to store smtp credentials correctly.

Add new attribute `smtp_parameters` to `IntegrationsState` class to store a `SmtpParameters` instance.

Add new function `_init_smtp` in `PaasCharm` class to handle smtp integration initialization.

Add a couple of unit tests for all frameworks.

### Checklist


src/paas_charm/charm.py
```diff
@@ -481,6 +514,12 @@ def _create_charm_state(self) -> CharmState:
                 "service_name": self.app.name,
                 "endpoint": f"{self._tracing.get_endpoint(protocol='otlp_http')}",
             }
+        smtp_relation_data = None
+        if self._smtp and (smtp_data := self._smtp.get_relation_data()):
+            smtp_relation_data = smtp_data.to_relation_data()
+            if password_id := smtp_relation_data.get("password_id"):
+                secret = self.model.get_secret(id=password_id)
+                smtp_relation_data["password"] = secret.get_content()
```
gregory-schiano, line 522: This should be provided by the library, when you're doing "get_relation_data" the library should extract the password value, and directly provide it in the pydantic object.
I think we should push that change in the library

examples/go/go_app/main.go
```diff
@@ -98,8 +98,8 @@ func main() {
 	if err := service.SubOperation(ctx); err != nil {
 		panic(err)
 	}
- 
-  	metricsPort, found := os.LookupEnv("APP_METRICS_PORT")
+
+	metricsPort, found := os.LookupEnv("APP_METRICS_PORT")
```
alithethird, line 102: line 101 had an empty space.
line 102 was using spaces instead of tabs. Rest of the file uses tabs.

src/paas_charm/charm.py
```diff
@@ -54,6 +54,15 @@
...
 
+try:
+    # pylint: disable=ungrouped-imports
+    from charms.smtp_integrator.v0.smtp import SmtpRequires
+except ImportError:
+    logger.exception(
+        "Missing charm library, please run "
+        "`charmcraft fetch-lib charms.smtp_integrator.v0.smtp`"
+    )
+
```
alithethird, line 65: Doing `logger.exception` might be confusing for users.
If we change it into `logger.warning` it will not print the traceback right? I think it would be better in terms of user experience if we do not print the traceback.

alithethird: I see that the [PR](https://github.com/canonical/paas-charm/pull/40) is open for it. Updated the new log for smtp library into warning.

src/paas_charm/app.py
```diff
@@ -322,6 +322,23 @@ def map_integrations_to_env(integrations: IntegrationsState, prefix: str = "") -
...
+                ("SMTP_USER", smtp.user),
+                ("SMTP_PASSWORD", smtp.password),
+                ("SMTP_AUTH_TYPE", smtp.auth_type),
+                ("SMTP_TRANSPORT_SECURITY", smtp.transport_security),
+                ("SMTP_DOMAIN", smtp.domain),
+                ("SMTP_SKIP_SSL_VERIFY", smtp.skip_ssl_verify),
+            )
+            if v is not None
+        )
+
```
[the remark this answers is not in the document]
alithethird: Planned to work on this in the design meeting but could not find time.

src/paas_charm/app.py
```diff
@@ -322,6 +322,23 @@ def map_integrations_to_env(integrations: IntegrationsState, prefix: str = "") -
...
+            (k, v)
+            for k, v in (
+                ("SMTP_HOST", smtp.host),
+                ("SMTP_PORT", str(smtp.port)),
+                ("SMTP_USER", smtp.user),
+                ("SMTP_PASSWORD", smtp.password),
+                ("SMTP_AUTH_TYPE", smtp.auth_type),
+                ("SMTP_TRANSPORT_SECURITY", smtp.transport_security),
+                ("SMTP_DOMAIN", smtp.domain),
+                ("SMTP_SKIP_SSL_VERIFY", smtp.skip_ssl_verify),
```
jdkandersson, line 337: Could we please take a look at the top few libraries in each framework we support and check that these environment variables are what they expect for SMTP?

tests/integration/flask/conftest.py
```diff
@@ -338,3 +344,73 @@ async def rabbitmq_k8s_integration_fixture(
...
+
+@pytest.fixture(scope="module", name="load_kube_config")
+def load_kube_config_fixture(pytestconfig: pytest.Config):
+    """Load kubernetes config file."""
+    kube_config = pytestconfig.getoption("--kube-config")
+    kubernetes.config.load_kube_config(config_file=kube_config)
+
+
+@pytest.fixture(scope="module")
+def mailcatcher(load_kube_config, ops_test: OpsTest):
```
jdkandersson, line 357: Could you provide some background here on your thinking please?

tests/unit/django/test_smtp.py
```diff
@@ -0,0 +1,81 @@
...
+    "auth_type": "none",
+    "domain": "example.com",
+    "host": "test-ip",
+    "port": "1025",
+    "skip_ssl_verify": "False",
+    "transport_security": "none",
+}
+
+
+def test_smtp_relation(harness: Harness, request: pytest.FixtureRequest):
```
jdkandersson, line 22: Could we do these as parametrised tests for each of the frameworks? Rather than repeat them a lot?

src/paas_charm/charm_state.py
```diff
@@ -402,8 +443,8 @@ class TempoParameters(BaseModel):
         service_name: Tempo service name for the workload.
     """
 
-    endpoint: str | None = None
-    service_name: str | None = None
+    endpoint: str = Field(alias="endpoint")
+    service_name: str = Field(alias="service_name")
```
alithethird, line 447: These should have been required all along.

If it's optional it doesn't fail when there is wrong integration data.

tests/integration/flask/conftest.py
alithethird: We need a local email server to be able to test properly. this one gets deployed into a kubernetes pod. I stole this from Weiis from penpot repo

examples/go/go_app/main.go
```diff
@@ -40,6 +43,87 @@ func (h mainHandler) serveHelloWorld(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Panic(err)
+	}
+
+	err = m.Close()
+	if err != nil {
+		log.Panic(err)
+	}
+
+	c.Quit()
```
javierdelapuente, line 122: maybe use defer

examples/go/go_app/main.go
```diff
@@ -40,6 +43,87 @@ func (h mainHandler) serveHelloWorld(w http.ResponseWriter, r *http.Request) {
...
+			InsecureSkipVerify: true,
+			ServerName:         smtp_host,
+		}
+		c.StartTLS(tlsconfig)
+	}
+
+	// Auth
+	if smtp_transport_security == "tls" {
+		if err = c.Auth(auth); err != nil {
+			log.Panic(err)
```
javierdelapuente, line 93: does panic crash the websever? we may want 500?

src/paas_charm/charm_state.py
```diff
@@ -483,6 +512,92 @@ def validate_signing_certificate_exists(cls, certs: str, _: ValidationInfo) -> s
...
+        """Turn auth_type type into None if its "none".
+
+        Args:
+            auth_type: Authentication type.
+
+        Returns:
+            The validated Authentication type.
+        """
+        if auth_type == AuthType.NONE:
+            return None
```
javierdelapuente, line 580: I do not understand what we are doing in here?

tests/unit/general/test_integrations.py
```diff
@@ -0,0 +1,867 @@
...
+            )
+    else:
+        relation_parameters = generate_relation_parameters(
+            relation_data, relation_parameter_type, accept_empty
+        )
+        assert isinstance(relation_parameters, expected_type)
+
+
+def _test_integrations_state_build_parameters():
+    mock_relations: dict[str, str] = {
```
javierdelapuente, line 376: maybe not mock in the name?

src/paas_charm/charm_state.py
alithethird: If we don't do this we have to check for it when generating environment variables from it. I think handling this in here is the cleanest way so we don't have to import this class into other places in the code.

alithethird: If we don't turn it into None, we either create a env variable with the value `none` or we check the `auth_type` to see if its not `none` then create the enb variable

jdkandersson (approved):

tests/unit/general/test_integrations.py
```diff
@@ -0,0 +1,867 @@
...
+def test_missing_required_other_integrations(
+    mock_charm, mock_requires, mock_charm_state, expected
+):
+    """
+    arrange: Get the mock charm, requires and charm state.
+    act: Run the _missing_required_other_integrations function.
+    assert: integration name should be in the result only when integration is required
+     and the parameters for that integration not generated.
+    """
+    expected = paas_charm.charm.PaasCharm._missing_required_other_integrations(
```
javierdelapuente, line 780: I think we should strive to test closer to the public interfaces of the classes. 

Any change of code in `_missing_required_other_integrations` (or moving code outside of that function), without changing the functionality, will probably break this test.

javierdelapuente (approved):
Looks good to me.

Minor comments to check.

PR #36, opened 2025-02-03
Relation parameter refactor

alithethird:
Applicable spec: <link>

### Overview
Updated how the charm creates relation parameter.

### Rationale
The currrent way we generate relation parameter classes from relation datas is almost identical but we are rewriting the logic for `s3` and `saml` relations. 

With the upcoming relations this rewriting of the  logic will increase. To prevent this I propose moving the logic into a function to be able to reuse it.

### Checklist


src/paas_charm/charm_state.py
```diff
@@ -241,35 +227,10 @@ def build(  # pylint: disable=too-many-arguments
...
-                saml_parameters = SamlParameters(**saml_relation_data)
-            except ValidationError as exc:
-                error_message = build_validation_error_message(exc)
-                raise CharmConfigInvalidError(
-                    f"Invalid Saml configuration: {error_message}"
-                ) from exc
-        else:
-            saml_parameters = None
+        s3_parameters, saml_parameters = list(
+            collect_relation_parameters(s3_connection_info, saml_relation_data)
```
javierdelapuente, line 232: Why not just collapse the `collect_relation_parameters` in here to make it more explicit, that is, just:
```
s3_parameters = generate_relation_parameters(s3_connection_info, S3Parameters)
saml_parameters = generate_saml_relation_parameters(saml_relation_data, SamlParameters)
```

src/paas_charm/charm_state.py
```diff
@@ -289,6 +250,92 @@ def build(  # pylint: disable=too-many-arguments
         )
 
 
+def generate_saml_relation_parameters(
```
javierdelapuente, line 253: cannot be generate_relation_parameters used for this case too?

alithethird: For saml the empty relation should be supported but for other relations empty should not be supported.

src/paas_charm/charm_state.py
alithethird: Hmm, yes this might be better but with the open PR's there is at least 2 more relations to add here. Using the `collect_relation_parameters` function makes this a bit cleaner I think.

src/paas_charm/charm_state.py
javierdelapuente: can't we unify them? I mean, behave the same way if possible? 

Besides, would it be possible to use a TypeVar for the "parameterized function"? Maybe there is another better way, but at least something like (I haven't tested but should work):
```
RelationParam = TypeVar("RelationParam", SamlParameter, S3Parameter)

def your_function(new_type: Type[RelationParam]) -> RelationParam:
    return new_type()
```

alithethird: To be honest I don't know why we need it to support empty relation but I know we specifically check for this in the unit tests. We also check for Saml to not support empty relation.

alithethird: For the other thing we need to cast the input dict into a specific parameter type. I don't understand how can we use the example code you shared.

javierdelapuente: I mean somethign like this:
```
RelationParam = TypeVar("RelationParam", SamlParameter, S3Parameter)

def generate_relation_parameters(
    relation_data: dict[str, str] | None,
    parameter_type: Type[RelationParam],
) -> "RelationParam | None":
    if not relation_data:
        return None
    try:
        return parameter_type(**relation_data)
    except ValidationError as exc:
        error_message = build_validation_error_message(exc)
        raise CharmConfigInvalidError(
            f"Invalid {parameter_type.__name__} configuration: {error_message}"
        ) from exc
```

alithethird: Oh, that is much nicer! Thanks a lot for teaching me this. I will add it immediately :)

alithethird: Can't get it to work. Linter really doesn't like this.

javierdelapuente: I just tried this and it works:
```
RelationParam = TypeVar("RelationParam", "SamlParameters", "S3Parameters")

def generate_relation_parameters(
    relation_data: dict[str, str] | typing.MutableMapping[str, str] | None,
    parameter_type: Type[RelationParam],
    support_empty: bool = False,
) -> "RelationParam | None":
    """Generate relation parameter class from relation data.

    Args:
        relation_data: Relation data.
        parameter_type: Parameter type to use.
        support_empty: Support empty relation data.

    Return:
        Parameter instance created.

    Raises:
        CharmConfigInvalidError: If some parameter in invalid.
    """
    if not support_empty and not relation_data:
        return None
    if relation_data is None:
        return None

    try:
        return parameter_type.parse_obj(relation_data)
    except ValidationError as exc:
        error_message = build_validation_error_message(exc)
        raise CharmConfigInvalidError(
            f"Invalid {parameter_type.__name__} configuration: {error_message}"
        ) from exc
```

alithethird: Awesome! Thanks.

alithethird: [note]
something
[/note]

alithethird: updated PR

src/paas_charm/charm_state.py
```diff
@@ -289,6 +252,55 @@ def build(  # pylint: disable=too-many-arguments
...
 
 
+RelationParam = TypeVar("RelationParam", "SamlParameters", "S3Parameters")
+
+
+def generate_relation_parameters(
+    relation_data: dict[str, str] | typing.MutableMapping[str, str] | None,
+    parameter_type: Type[RelationParam],
+    support_empty: bool = False,
+) -> "RelationParam | None":
```
javierdelapuente, line 262: No need for "

javierdelapuente (approved):

src/paas_charm/charm_state.py
alithethird: removed

jdkandersson:
Please take a look at the description and add more details. I.e., what the refactor actually is and why it is needed 🙂

src/paas_charm/charm_state.py
```diff
@@ -241,35 +225,14 @@ def build(  # pylint: disable=too-many-arguments
...
-                ) from exc
-        else:
-            saml_parameters = None
+        s3_parameters = typing.cast(
+            S3Parameters | None, generate_relation_parameters(s3_connection_info, S3Parameters)
+        )
+        saml_parameters = typing.cast(
+            SamlParameters | None,
+            generate_relation_parameters(saml_relation_data, SamlParameters, True),
+        )
```
jdkandersson, line 235: I think I don't think the type casting is necessary here since a `TypeVar` is being used to define the argument type?

jdkandersson (approved):

PR #41, opened 2025-02-10
Chore(refactor): Put all integration reqs in one class

alithethird:
### Overview

Refactoring the way integration data passed onto the `CharmState`  class. 

### Rationale

With the latest integration additions, the `Charmstate.from_charm` function requires more and more arguments. 

Instead of adding new arguments we can create a class that collects all the integration requirer objects and pass an instance of this class to the `Charmstate.from_charm`  function.  

### Module Changes

Add new class called `IntegrationRequirers`.  This class holds all the integration requirer objects.

The necessary integration data to build `IntegrationState` class is generated in the `Charmstate.from_charm` function by using the requrier objects stored in the `IntegrationRequirers` class.

### Checklist


jdkandersson (approved):

src/paas_charm/charm_state.py
```diff
@@ -10,15 +10,35 @@
...
 from paas_charm.secret_storage import KeySecretStorage
 from paas_charm.utils import build_validation_error_message
 
 logger = logging.getLogger(__name__)
 
+try:
+    # the import is used for type hinting
+    # pylint: disable=ungrouped-imports
+    # pylint: disable=unused-import
+    from charms.data_platform_libs.v0.s3 import S3Requirer
```
javierdelapuente, line 28: we are asking for trouble in here. This may just work because the typing is done with quotes, like  `s3: "S3Requirer | None" = None`. If it was `s3: S3Requirer | None = None` this will fail at runtime.

src/paas_charm/charm.py
```diff
@@ -490,12 +486,14 @@ def _create_charm_state(self) -> CharmState:
             framework=self._framework_name,
             framework_config=self.get_framework_config(),
             secret_storage=self._secret_storage,
-            database_requirers=self._database_requirers,
-            redis_uri=self._redis.url if self._redis is not None else None,
-            s3_connection_info=self._s3.get_s3_connection_info() if self._s3 else None,
-            saml_relation_data=saml_relation_data,
-            rabbitmq_uri=self._rabbitmq.rabbitmq_uri() if self._rabbitmq else None,
-            tracing_requirer=self._tracing if self._tracing is not None else None,
+            integration_requirers=IntegrationRequirers(
```
javierdelapuente, line 489: I would put the creation of this object in a private method of the charm, get the relation data (instead of passing the library) when possible. That way the mocks in `test_charm_state.py` can also be deleted and the code is already tested in `flask/test_charm.py::test_integrations_wiring`.

src/paas_charm/charm_state.py
alithethird: But since we use quotes it should not fail at runtime right? That was my thinking.

src/paas_charm/charm.py
javierdelapuente: Out of curiosity, any reason for not doing it in the charm? From my point of view, the sooner we stop the use of the requirer libraries the better.

src/paas_charm/charm_state.py
javierdelapuente: I was just thinking that I see them, and check that the tests pass without them, I would probably remove them...

alithethird: Well, yes. All  the tests pass but the linter is not happy about it. I don't know which is better:
- Adding the includes
- Adding ignore comments for mypy.
wdyt @jdkandersson ?

src/paas_charm/charm.py
alithethird: Wanted to keep the integration logic in `CharmState` and not leak it into the `PaasCharm` class

src/paas_charm/charm_state.py
javierdelapuente: At least I would add a comment, that those quotes are just a trick for not crashing and have the linter happy (I think of that as a "pylint" exception.

src/paas_charm/charm.py
javierdelapuente: Ok, let's put it that way (I was thinking the opposite way, we are leaking charm things into the charm_state when they can be removed beforehand)

javierdelapuente (approved):

PR #42, opened 2025-02-11
Feat(Block on missing config)

alithethird:
### Overview
Some applications require certain configuration options to function correctly. With this PR when a non-optional config option is missing, the charm will go into the blocked state with a message telling which config options are missing.

### Rationale

Some applications require certain config options to run and would crash if these options are not provided. To prevent this, we need to add an optional field to the config options and block the charm if a config option marked as non-optional has not been provided with a message that explains which config options are missing.

To stay backward compatible the default value for the optional field will be true.

Since it doesn’t make sense for config options that are non-optional to have a default value, charmcraft will look for this condition at pack time and error out.

### Module Changes

2 new functions in `paas_charm.charm_state.py` are added to create the `AppConfig` class from the list of configuration options:
- `app_config_factory`
- `_create_configuration_attribute`

`paas_charm.utils.build_validation_error_message` function is changed. Now it returns a tuple of 2 strings.

The tuple consists of:
-  A list of error fields from a pydantic exception to print at charm status.
-  A detailed list of error fields from a pydantic exception with their error
             messages. Best fit for printing error logs.

A new function in `paas_charm.utils.py`:
- `is_user_defined_config`: This function checks whether a config option is user-defined.

Add a new folder to unit tests to check functions not affected by the framework.
Add a new folder to integration tests to check functionality in every framework by parametrizing the framework. This way we only need to write 1 function to test functionality in every framework.

### Checklist


tests/unit/flask/test_charm.py
```diff
@@ -348,6 +353,107 @@ def test_invalid_config(harness: Harness):
...
+    harness.charm.get_framework_config()
+    is_blocked = isinstance(harness.model.unit.status, ops.model.BlockedStatus)
+
+    print(yaml_dict["config"]["options"])
+    with open(charm_dir / "charmcraft.yaml", "w") as file:
+        file.write(original_file)
+
+    assert is_blocked
+
+
```
alithethird, line 456: This tests are working when they are run explicitly, they don't work when all unit tests are run at the same time.

Reading and updating the yaml file will be a fixture in the following PR's. Currently just trying to get the test working.

tests/unit/flask/test_charm.py
```diff
@@ -348,6 +353,107 @@ def test_invalid_config(harness: Harness):
...
+    config_file = charm_dir / "charmcraft.yaml"
+    if config_file.exists():
+        yaml_dict = yaml.safe_load(config_file.read_text())
+    new_yaml = copy.deepcopy(yaml_dict)
+
+    original_file = ""
+
+    with open(charm_dir / "charmcraft.yaml", "r") as file:
+        original_file = file.read()
```
alithethird, line 440: I am explicitly reading and writing the file because I discovered dumping the yaml is changing some field's order.

src/paas_charm/utils.py
```diff
@@ -103,3 +103,24 @@ def config_get_with_secret(
...
+
+    Returns:
+        Returns the list of missing non-optional config options,
+        if none exists returns None.
+    """
+    metadata = _config_metadata(pathlib.Path(os.getcwd()))
+    missing_options = []
+    for option in metadata["options"]:
+        if metadata["options"][option].get("optional") is False:
+            if not charm.config.get(option):
```
weiiwang01, line 122:
```suggestion
            if charm.config.get(option) is None:
```
I think using `is None` would be better here since some configurations may contain `0` or `""`, which are falsey.

src/paas_charm/utils.py
```diff
@@ -103,3 +103,24 @@ def config_get_with_secret(
...
+                missing_options.append(option)
+    if missing_options:
+        return missing_options
+    return None
```
weiiwang01, line 126:
```suggestion
    return missing_options
```
An empty list is natural for representing the no missing options.

weiiwang01 (approved):

jdkandersson:
Is this fully ready? At the interface PR stage, this should be marked as draft please

src/paas_charm/utils.py
```diff
@@ -103,3 +103,24 @@ def config_get_with_secret(
     if secret_id is None:
         return None
     return charm.model.get_secret(id=typing.cast(str, secret_id))
+
+
+def get_missing_non_optional_configs(charm: ops.CharmBase) -> list[str] | None:
```
jdkandersson, line 108: What is the reason for returning `None`? Why not just return an empty collection?

src/paas_charm/utils.py
```diff
@@ -103,3 +103,24 @@ def config_get_with_secret(
...
+
+    Args:
+        charm: The charm instance.
+
+    Returns:
+        Returns the list of missing non-optional config options,
+        if none exists returns None.
+    """
+    metadata = _config_metadata(pathlib.Path(os.getcwd()))
+    missing_options = []
```
jdkandersson, line 119: I think this implementation could be cleaner

```python
options = metadata["options"]
non_optional_options = (option for option in options if metadata["options"][option].get("optional") is False)
return [option for option in non_optional_options if charm.config.get(option) is None]
```

tests/unit/flask/test_charm.py
```diff
@@ -348,6 +354,79 @@ def test_invalid_config(harness: Harness):
...
+    harness.charm.get_framework_config()
+
+    if missing_configs:
+        assert isinstance(harness.model.unit.status, ops.model.BlockedStatus)
+        for config in missing_configs:
+            assert config in str(harness.model.unit.status.message)
+    else:
+        assert isinstance(harness.model.unit.status, ops.model.ActiveStatus)
+
+    paas_charm.utils._config_metadata = _config_metadata
```
alithethird, line 427: If i don't have this line, the mock stays there and fails other tests.

src/paas_charm/charm_state.py
```diff
@@ -126,6 +138,14 @@ def from_charm(  # pylint: disable=too-many-arguments
             k: v for k, v in app_config.items() if k not in framework_config.dict().keys()
         }
 
+        app_config_class = app_config_factory(framework)
+        try:
+            app_config_class(**app_config)
+        except ValidationError as exc:
+            error_message = build_validation_error_message(exc, underscore_to_dash=True)
+            logger.error(error_message[1])
+            raise CharmConfigInvalidError(error_message[0]) from exc
```
alithethird, line 147: If I unpack the tuple in line 145, I get MyPy `too-many-locals` warning.

src/paas_charm/charm_state.py
```diff
@@ -458,3 +479,64 @@ def validate_signing_certificate_exists(cls, certs: str, _: ValidationInfo) -> s
...
+        case "float":
+            config_type = float  # type: ignore[assignment]
+        case "string":
+            config_type = str  # type: ignore[assignment]
+        case "secret":
+            config_type = dict  # type: ignore[assignment]
+        case _:
+            raise ValueError(f"Invalid option type: {config_type_str}.")
+
+    type_tuple = (config_type, ...)
```
alithethird, line 517: Here, the left side of the tuple is the attribute type and the right side is the default value.

Using `None` as the default value automatically makes the attribute optional so I have to use `ellipsis` as the default value for non-optional attributes.

src/paas_charm/utils.py
```diff
@@ -15,27 +14,56 @@
...
+            error_field_str += f"{err[0]}, "
+        error_field_log_str += f"\n\t- {err[0]}: {err[1]}"
+
+    message_str = ""
+    if missing_options_str:
+        message_str += f"missing options: {missing_options_str[:-2]} "
+    if error_field_str:
+        message_str += f"invalid options: {error_field_str[:-2]}"
+
+    return (message_str, error_field_log_str)
```
jdkandersson, line 66:
```
missing_fields = {}
invalid_fields = {}

for err in exc.errors():
    if "required" in err["msg"].lower():
        missing_fields[err["loc"]] = err["msg"]
    if "invalid" in err["msg"].lower():
        invalid_fields[err["loc"]] = err["msg"]

message_str_missing = f"missing options: ', '.join(missing_fields)" if missing_fields else ""
message_str_invalid = f"invalid options: ', '.join(invalid_fields)" if invalid_fields else ""
message_str = f"{message_str_missing}{', ' if message_str_missing and message_str_invalid else ''}{message_str_invalid}"

"""
invalid configuration:
- <field name>: <issue>
"""

log_message_lines = (f"- {key}: {value}" for key, value in itertools.chain(missing_fields.items(), invalid_fields.items()))
log_message = f"invalid configuration:\n{'\n'.join(log_message_lines)}"
```

.github/workflows/integration_test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       extra-arguments: -x --localstack-address 172.17.0.1
       pre-run-script: localstack-installation.sh
       charmcraft-channel: latest/edge
-      modules: '["test_charm.py", "test_cos.py", "test_database.py", "test_db_migration.py", "test_django.py", "test_django_integrations.py", "test_fastapi.py", "test_go.py", "test_integrations.py", "test_proxy.py", "test_workers.py", "test_tracing.py"]'
+      modules: '["test_charm.py", "test_cos.py", "test_database.py", "test_db_migration.py", "test_django.py", "test_django_integrations.py", "test_fastapi.py", "test_go.py", "test_integrations.py", "test_proxy.py", "test_workers.py", "test_tracing.py", "test_charm_state.py"]'
```
jdkandersson, line 17: Instead of `test_charm_state` perhaps we do `test_config` or `test_user_input`

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 # See LICENSE file for licensing details.
 [project]
 name = "paas-charm"
-version = "1.2.3"
+version = "1.2.4"
```
jdkandersson, line 5: Let's fix the version please

src/paas_charm/charm_state.py
```diff
@@ -458,3 +483,61 @@ def validate_signing_certificate_exists(cls, certs: str, _: ValidationInfo) -> s
...
+        case "string":
+            config_type = str
+        case "secret":
+            config_type = dict
+        case _:
+            raise ValueError(f"Invalid option type: {config_type_str}.")
+
+    type_tuple: tuple = (config_type, Field())
+    if optional:
+        type_tuple = (config_type | None, None)
```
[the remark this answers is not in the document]
alithethird: Mypy still wants `:tuple`

```python
if optional:
    type_tuple: tuple = (config_type | None, None)
else:
    type_tuple = (config_type, Field())
```

src/paas_charm/utils.py
```diff
@@ -13,29 +13,74 @@
...
     Args:
         exc: ValidationError exception instance.
         prefix: Prefix to append to the error field names.
         underscore_to_dash: Replace underscores to dashes in the error field names.
 
     Returns:
-        The curated list of error fields ready to be used in an error message.
+        The tuple of 2 strings first for charm status, second for logging.
     """
     error_fields_unique = set(
```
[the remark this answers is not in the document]
alithethird: We can not use `itertools.chain.from_iterable` because it flattens tuples. The resultant `fields` would not be a chain of tuples but a chain of strings.

src/paas_charm/charm_state.py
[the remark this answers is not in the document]
jdkandersson: What  is the mypy error please?

weiiwang01 (approved):

tests/integration/general/test_config.py
```diff
@@ -0,0 +1,84 @@
...
+        assert invalid_config in blocked_app.status_message
+
+    await blocked_app.set_config(set_first_configs)
+    await model.wait_for_idle(apps=[blocked_app.name], status="blocked", timeout=300)
+    for invalid_config in rest_of_the_configs:
+        assert invalid_config in blocked_app.status_message
+    for config in set_first_configs.keys():
+        assert config not in blocked_app.status_message
+
+    await blocked_app.set_config(set_the_rest_of_configs)
```
jdkandersson, line 81: This is also confusing, is it a set? Or what is this?

Why does the order of the config options matter?

alithethird: `set` here is used as a verb. I did not know what else to call it.

The idea here is that we set some of the missing options to see that the charm is still blocked and the status message changes reflecting only some of the missing options are set.

alithethird: Should I call it `first_optional_config` and the rest of the configs `remaining_optional_configs` ?

jdkandersson (approved):
Please remember to update the charmcraft docs

PR #66, opened 2025-04-16
chore: refactor s3 integration & use S3 environment variable mapper

yanksyoon:
Applicable spec: <link>

### Overview

- Minor chores (alphabetical import, development artifact gitignore)
- Introduce S3 integration lib wrapper (to_relation_data method that returns a Pydantic object)
- Introduce default S3EnvironmentMapper that can be overriden

### Rationale

- Later on, some custom environment mappers are required for SpringBoot support. We will refactor to introduce EnvironmentVariableMappers which will be overriden with custom EnvironmentVariableMapper for each required integrations.

### Checklist

- [ ] The [changelog](../docs/changelog.md) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)

- No user relevant changes, all interfaces are kept the same.


src/paas_charm/s3.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        connection_info = self.get_s3_connection_info()
+        if not connection_info:
+            return None
+        try:
+            return S3RelationData.model_validate(connection_info)
+        except ValidationError as exc:
+            error_messages = build_validation_error_message(exc, underscore_to_dash=True)
+            logger.error(error_messages.long)
+            raise CharmConfigInvalidError(
```
gregory-schiano, line 79: I think we could have a specific type of error for the wrapped libraries

gregory-schiano (commented):
I like the end results, I think once everything is moved to this pattern, it'll be easier to add extra relations :)

src/paas_charm/app.py
```diff
@@ -17,6 +17,16 @@
 
 logger = logging.getLogger(__name__)
 
+try:
+    # the import is used for type hinting
```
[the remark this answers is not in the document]
yanksyoon: Added `TYPE_CHECKING` but also persisted the try-catch. The import can fail if the charm libs have not been fetched.
Currently, all places that uses charm libs have try-catch to prevent runtime error, should we centralize it?

src/paas_charm/s3.py
yanksyoon: Wrapped it in `InvalidS3RelationError` - I believe it is better to not catch this and raise an error since it's is merely a lib wrapper! :)

src/paas_charm/app.py
```diff
@@ -69,10 +79,60 @@ def should_run_scheduler(self) -> bool:
         return unit_id == "0"
 
 
+class S3EnvironmentMapper:  # pylint: disable=too-few-public-methods
```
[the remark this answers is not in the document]
yanksyoon: Since the default environment mapper is embedded in the App class as follows:
```
class App:
    s3_environ_mapper = S3EnvironmentMapper
```

it is clearer to have a static method inside the S3EnvironmentMapper than passing a function/method. The reason is the function/method would require passing the initial `cls` argument, making it look like the following:
```
def default_s3_mapper_func(cls, relation_data):
    ...

class App:
    s3_environ_mapper = default_s3_mapper_func
```
which could be easier to make mistakes in not recognizing the `cls` initial parameter when injecting/overriding it from the child classes.

Another reason is for later on - when environment mappers from child classes are to be introduced, it is easier to have a generic interface using ABCs that can inherit from the existing mapper class.

What do you think about this approach?

src/paas_charm/app.py
```diff
@@ -17,6 +17,17 @@
 
 logger = logging.getLogger(__name__)
 
+if TYPE_CHECKING:
+    try:
```
yanksyoon, line 21: remove try-catch

src/paas_charm/charm_state.py
```diff
@@ -156,42 +155,50 @@ def from_charm(  # pylint: disable=too-many-arguments
...
-                if (
-                    integration_requirers.smtp
-                    and (smtp_data := integration_requirers.smtp.get_relation_data())
-                )
-                else None
-            ),
-        )
+        # 2025/04/15 When done ejecting IntegrationParameters, we should remove the build function
+        # and just wrap the PydanticObjects from IntegrationRequirer libs into the
+        # IntegrationState, without the build function. See integration_requirers.s3.
```
cbartz, line 160: Could we add this to the jira backlog? Usually these kind of comments live forever in the code and confuse future developers :D

tests/unit/general/test_app.py
```diff
@@ -0,0 +1,66 @@
...
+
+@pytest.mark.parametrize(
+    "relation_data, expected_env",
+    [
+        pytest.param(None, {}, id="No relation data"),
+        pytest.param(
+            S3RelationData.model_construct(access_key="key", secret_key="key", bucket="bucket"),
+            {
+                "S3_ACCESS_KEY": "key",
+                "S3_SECRET_KEY": "key",
```
cbartz, line 23: nit: maybe use another value as for the access key

tests/unit/general/test_s3.py
```diff
@@ -0,0 +1,105 @@
...
+                        "header2:value2",
+                    ]
+                ),
+            ),
+            id="Maximum data",
+        ),
+    ],
+)
+def test_requirer_to_relation_data(monkeypatch, relation_data, expected):
+    s3_requirer = PaaSS3Requirer(MagicMock(), "test-s3-integration", "test-bucket")
```
cbartz, line 79: nit: could you add the "arrange act assert" doc strings to the tests?

cbartz (approved):
LGTM, have some nits

src/paas_charm/app.py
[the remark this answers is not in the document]
yanksyoon: Updated to use staticmethod, thank you :)

src/paas_charm/charm.py
```diff
@@ -32,7 +32,7 @@
 # if new optional libs are not fetched, as it will not be backwards compatible.
 try:
     # pylint: disable=ungrouped-imports
-    from charms.data_platform_libs.v0.s3 import S3Requirer
+    from paas_charm.s3 import PaaSS3Requirer
 except ImportError:
     logger.warning(
         "Missing charm library, please run `charmcraft fetch-lib charms.data_platform_libs.v0.s3`"
```
alithethird, line 38: Should we move the try-catch block to the `paas_charm.s3` module?

yanksyoon: https://github.com/canonical/paas-charm/pull/66/files#diff-03aeeeb76b08f6d45e96c954dd34f3643656866e785458c73e29a7656472dd8cR8-R10

I've added a comment on the S3 lib on the reason why I'm letting the error fallthrough :)

src/paas_charm/s3.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Provide a wrapper around S3 integration lib."""
+import logging
+from typing import Optional
+
+# The import may fail if optional libs are not fetched. Let it fall through
+# and let the caller (charm.py) handle it, to make this wrapper act like the
+# native lib module.
```
alithethird, line 10: I see your point but I feel like we are going in the direction where integration wrappers we write will handle everything regarding the integration.
So handling the import fail here would be better in my opinion. 

It will also clean up some of the clutter on the `app.py` which is always nice.

tests/unit/flask/test_charm_state.py
```diff
@@ -132,29 +132,7 @@ def test_s3_integration(s3_connection_info, expected_s3_parameters):
         ),
     )
     assert charm_state.integrations
-    assert charm_state.integrations.s3_parameters == expected_s3_parameters
-
-
-def test_s3_integration_raises():
```
alithethird, line 138: Can we include a test for the invalid data?

tests/unit/flask/test_flask_app.py
```diff
@@ -15,7 +15,7 @@
 from paas_charm._gunicorn.workload_config import create_workload_config
 from paas_charm._gunicorn.wsgi_app import WsgiApp
 from paas_charm.app import map_integrations_to_env
-from paas_charm.charm_state import CharmState, IntegrationsState, S3Parameters
+from paas_charm.charm_state import CharmState, IntegrationsState, S3RelationData
```
alithethird, line 18: Unused import?

src/paas_charm/s3.py
yanksyoon: Yes - I agree, perhaps we could follow up on this discussion here: https://github.com/canonical/paas-charm/pull/66#discussion_r2050438361

I think we can create a ticket on following up on this item since this impacts the design of `app.py`. There are import checking code that need to be modified as well with this update.

tests/unit/flask/test_charm_state.py
yanksyoon: https://github.com/canonical/paas-charm/pull/66/files/25f78bedb3ccc8078d57a742e821b3d48571d5ca#diff-ba57531283b9fc16630dac3421730548b7138d428fb41693d83b159b90e82293R87-R105

Invalid data tests have been moved here

tests/unit/flask/test_flask_app.py
yanksyoon: Surprised that the tests don't catch this :o

tests/unit/general/test_s3.py
alithethird: +1

src/paas_charm/charm_state.py
yanksyoon: Added https://warthogs.atlassian.net/browse/ISD-3362

tests/unit/flask/test_charm_state.py
alithethird: Oh, awesome!

src/paas_charm/app.py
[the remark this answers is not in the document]
alithethird: I feel like we are on a path to create wrappers for all integrations and let the wrappers handle everything related to the integrations, and the try-catch blocks should be the integrations responsibility. Moving them into the wrapper would be better in my opinion.

alithethird (approved):
Overall, great PR!

Thanks a lot for your work:scientist: 

I only had minor comments. Also, sorry for the comments being separate from this review. I just started using the VSCode PR extension and did the commenting wrong, oops :laughing:

src/paas_charm/app.py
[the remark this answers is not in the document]
yanksyoon: I've added https://warthogs.atlassian.net/browse/ISD-3362 to address this later on :)

PR #68, opened 2025-04-17
feat: initial saml wrapper

yanksyoon:
Applicable spec: <link>

### Overview

- Refactor SAML parameters to SAML relation data
- Introduce SAML environment mapper

### Rationale

- To introduce environment variable mappers, follow up to: https://github.com/canonical/paas-charm/pull/66

### Checklist

- [ ] The documentation for charmhub is updated
- [ ] The [changelog](../docs/changelog.md) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)

No user relevant changes


src/paas_charm/saml.py
gregory-schiano, line 1: I'm not convinced we need a wrapper for SAML:
- Either these properties/validators aren't specific to 12f, in which case I'm wondering if we shouldn't add them to the SAML library
- Or these transformations / validation could be done in the envvar mapper. We're not really altering the SAML pydantic model

WDYT @javierdelapuente ?

yanksyoon: I created this validator to keep the behavior the same but it does seem that the transformations could be done in the env mappers. But then, do we change the existing behavior of the validation errors?

tests/unit/general/test_saml.py
```diff
@@ -0,0 +1,137 @@
...
+                        response_url="http://test-response-url.test",
+                    ),
+                ],
+            ),
+            id="maximum relation data",
+        ),
+    ],
+)
+def test_paas_saml_requirer_to_relation_data(relation_data, expected_relation_data):
+    """Test PaaSSAMLRequirer to relation data conversion."""
```
cbartz, line 93: nit: can you use our arrange format for doc string in the tests?

cbartz (approved):

tests/unit/general/test_app.py
```diff
@@ -0,0 +1,72 @@
...
+                        response_url="http://response-url.test",
+                    ),
+                ),
+            ),
+            {
+                "SAML_ENTITY_ID": "test-entity-id",
+                "SAML_METADATA_URL": "http://test-metadata.url",
+                "SAML_SINGLE_SIGN_ON_REDIRECT_URL": "http://testing-redirect-url.test/",
+                "SAML_SIGNING_CERTIFICATE": "test-certificate-1,test-certificate-2",
+            },
```
alithethird, line 61: Can we also add a test for modified env mapper?

alithethird (approved):
Great PR again!

I was wondering should we keep the integration tests parametrized or put in different functions in different files?

I feel like parametrization would make writing tests for new integrations a lot easier.

tests/unit/general/test_app.py
yanksyoon: I think that test should come in when the modification (override) happens - i.e. when this override mapper is introduced with the SpringBoot Application :)

PR #160, opened 2025-09-05
docs: Explain the opinionated nature of the tooling

erinecon:
Applicable ticket: ISD-3682

### Overview

Add an explanation document to describe the opinionated nature of the 12-factor tooling.

### Rationale

In UX sessions and user interviews, participants appreciated the "magic" nature of the tooling and the ability to set up rocks and charms with less effort and difficulty. However, they also wanted to peak behind the curtain and understand the "why" behind the tooling.

Similarly, participants noted the lack of clarity behind the opinionated nature of the tooling, and wished to understand what the opinions are and which ones can be overridden. 

### Juju Events Changes

None

### Module Changes

None

### Library Changes

None

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The RTD documentation is updated
- [X] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)
- [X] The [changelog](../docs/changelog.md) is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)

### DIFF
--- a/docs/explanation/opinionated-nature-tooling.rst
+++ b/docs/explanation/opinionated-nature-tooling.rst
@@ -1,0 +1,218 @@
+.. _explanation_opinionated_nature:
+
+Opinionated nature of the 12-factor tooling
+===========================================
+
+The native 12-factor framework support in Rockcraft and Charmcraft provides
+a streamlined approach to setting up a rock and charm for a web app, but this
+support contains strong and weak opinions about the project structure, the
+Ubuntu base, framework versions, and configurations. This document describes
+the various ways in which the 12-factor tooling is opinionated.
+
+.. note::
+
+    There are several components of 12-factor app rocks and charms that can
+    be configured or customized by the user. See :ref:`ref_supported_customization`
+    for a complete list.
+
+Project structure
+-----------------
+
+For Rockcraft to successfully initialize the rock for a web app, it inspects the
+directory where ``rockcraft init`` is run to search for specific files and global
+variables depending upon the selected framework. The following sub-sections
+detail the requirements for each of the supported frameworks.
+
+Django
+~~~~~~
+
+The project must be set up with the structure
+``project-name/project_name/project_name``, where ``project-name`` specifies the name
+of the Django project. The rock must be initialized in the ``project-name`` directory.
+
+There must be a ``project-name/requirements.txt`` file where Django is declared as
+a dependency. There must also be a ``project-name/project_name/project_name/wsgi.py`` file
+in the project.
+
+When ``rockcraft init`` is run, Rockcraft searches for these two files
+and expects to find a global variable named ``application`` in the ``wsgi.py`` file. 
+
+The initialization will fail if Rockcraft does not find the ``requirements.txt`` file,
+the ``wsgi.py`` file, or the global variable ``application``.
+
+Express
+~~~~~~~
+
+The Express app must reside in a dedicated ``app`` directory containing a ``package.json`` file.
+The ``package.json`` file must contain a ``scripts`` key, and this key must contain a ``start``
+script which defines how the server is run. Furthermore, the ``name`` key in the file must be
+defined and set to the app name.
+
+The rock must be initialized outside of the ``app`` directory.
+When ``rockcraft init`` is run, Rockcraft searches for the ``app/package.json`` file
+and expects to find both the ``scripts: start`` and ``name`` keys. 
+
+The initialization will fail if Rockcraft does not find the file and both keys.
+
+FastAPI
+~~~~~~~
+
+The FastAPI project must contain a ``requirements.txt`` file where ``fastapi`` is
+listed as a dependency. 
+The rock must be initialized in the same directory as the ``requirements.txt`` file.
+
+There must also be an ASGI app in the FastAPI project with the global variable ``app``.
+Rockcraft will search for different names and locations for this file:
+
+.. list-table::
+  :header-rows: 1
+  :widths: 5 10
+
+  * - Possible project file name
+    - Possible project file locations
+  * - ``app.py``
+    - Same directory as ``requirements.txt`` file, ``/app``, or ``/src``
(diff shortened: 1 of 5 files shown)


docs/explanation/opinionated-nature-tooling.rst
```diff
@@ -0,0 +1,218 @@
...
+an ``mvnw`` executable file in the same directory.
+If you use a ``build.gradle`` file, there must also exist a ``gradlew`` executable file
+in the same directory.
+The project must not contain both executable files, otherwise the rock initialization
+will fail.
+
+When ``rockcraft init`` is run, Rockcraft searches for either of these files and their
+corresponding executable files.
+The initialization will fail if Rockcraft does not find either file and its corresponding
+executable file, or if it finds both files or both executable files.
```
erinecon, line 137: @javierdelapuente did I understand https://github.com/canonical/rockcraft/blob/main/rockcraft/extensions/springboot.py correctly here?

docs/explanation/opinionated-nature-tooling.rst
```diff
@@ -0,0 +1,218 @@
...
+  * - Express
+    - ???
+  * - FastAPI
+    - ???
+  * - Flask
+    - ???
+  * - Go
+    - ???
+  * - Spring Boot
+    - ???
```
erinecon, line 162: @javierdelapuente and @alithethird , I would really appreciate your help determining the minimum supported versions for each of the frameworks.

docs/explanation/opinionated-nature-tooling.rst
```diff
@@ -0,0 +1,218 @@
...
+Available charm features
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some features of the charm are forbidden to add or modify, meaning that ``charmcraft pack``
+will fail if those features were changed. 
+
+For instance, storage cannot be added to a 12-factor app charm. It is required that the
+charm is stateless, meaning that any reusable files created by the app cannot be shared
+between other charms with which it is integrated in the Juju ecosystem. Juju doesn't
+support a shared file system.
```
erinecon, line 215: @javierdelapuente and @weiiwang01 , I'm feeling quite uncertain about what I described here. Could you check this section in particular and make sure it's accurate and makes sense?

Weii, Greg told me that you had written a spec covering features that are forbidden to add or modify in the charm. Do you happen to have a link for that spec, or could you point me to a resource with more information?

docs/explanation/opinionated-nature-tooling.rst
javierdelapuente: That's a good question. Any recent version should work, but I do not think we have set those explicitly (and of course, we could be broken if some future changes happen).

Also, in any case, for example, we are limited by the Python version for Django/FastAPI/Flask (for example, python 3.12 will work with certain versions of Django). Go should not be a problem and in the case of Spring Boot, we could limit ourselves to the versions included in devpack... 

This is something to think a bit more anyway...

docs/explanation/opinionated-nature-tooling.rst
```diff
@@ -0,0 +1,218 @@
...
+  * - Spring Boot
+    - ???
+
+Opinions related to the base
+----------------------------
+
+Depending on the Ubuntu base specified in the ``rockcraft.yaml`` file, opinions are imposed
+about the Python version used:
+
+* Specifying ``base: ubuntu@22.04`` will use Python 3.10
```
javierdelapuente, line 170: base or build-base, depending if it is a bare base

docs/explanation/opinionated-nature-tooling.rst
javierdelapuente: Juju support volumes. We do not allow them in the 12factor charmcraft extension. To use them the user cannot use the extension.

`reusable files created by the app cannot be shared between other charms`. Not sure what you want to mean with that sentence.

javierdelapuente (commented):
a few comments

docs/explanation/opinionated-nature-tooling.rst
```diff
@@ -0,0 +1,218 @@
...
+Go
+~~
+
+The Go project must contain a ``go.mod`` file. The rock must be initialized in the
+same directory as this file.
+
+When ``rockcraft init`` is run, Rockcraft searches for this file. The initialization will
+fail if Rockcraft does not find the file.
+
+Rockcraft expects to find a binary with the same name as the Go project. If the binary meets
```
alithethird, line 115: Rockcraft pack is building the binary itself, so saying here that `Rockcraft expects to find a binary` might be confusing.

docs/explanation/opinionated-nature-tooling.rst
alithethird: That seems about right. Thank you @erinecon

docs/explanation/opinionated-nature-tooling.rst
alithethird: I am thinking about testing the versions with a script maybe? @javierdelapuente 
do you have a more efficient idea?

alithethird (commented):
Looks pretty good! Just 1 comment.

docs/explanation/opinionated-nature-tooling.rst
javierdelapuente: Not sure, it is going to be difficult to keep up to date. Seeing for example [Django](https://docs.djangoproject.com/en/5.2/faq/install/#what-python-version-can-i-use-with-django), current one is 5.2, that requires python `3.10, 3.11, 3.12, 3.13`. 

Maybe we can open another PR and discuss it there to not block this one.

PR #280, opened 2026-05-25
fix: Include 404 to retry list

alithethird:
#### What this PR does

Adds 404 status to retry statuses.

#### Why we need it

Especially the Go port test sometimes gets flaky and the request needs to be retried. The fail status is 404 in this test but the retry logic did not include 404 status.

#### Checklist

- [ ] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I added or updated the documentation (if applicable)
- [ ] I updated `docs/changelog.md` with user-relevant changes
- [ ] I used AI to assist with preparing this PR
- [ ] I added or updated tests as needed (unit and integration)
- [ ] **If integration test modules are used:** I updated the workflow configuration  

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -7,6 +7,9 @@ on:
 
 jobs:
   integration-tests:
+    permissions:
+      packages: write
+      pull-requests: write
     strategy:
       matrix:
         juju-version: [3.6/stable]
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -38,7 +38,7 @@ def fixture_session_with_retry():
         read=5,
         other=5,
         backoff_factor=5,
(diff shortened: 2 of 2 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -7,6 +7,9 @@ on:
 
 jobs:
   integration-tests:
+    permissions:
+      packages: write
+      pull-requests: write
```
swetha1654, line 12: why do we need this?

swetha1654 (approved):

.github/workflows/integration_test.yaml
alithethird: The rocks can not be uploaded to ghcr.io if we don't have them :(

alithethird: They recently changed the default permissions

PR #282, opened 2026-05-26
Add support for the valkey_client interface

Thanhphan1147:
#### What this PR does

**! This PR also bumps ops[testing] to 3.7 because it's required by dpcharmlibs-interfaces **
- ops[testing] >= 3 removes the automatic peer relation data for the current unit in the databag, so this PR also removes a lot of `APP_PEER_FQDN` checks in the unit tests.
#### Checklist

- [x] **If integration test modules are used:** I updated the workflow configuration  


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 # See LICENSE file for licensing details.
 [project]
 name = "paas-charm"
-version = "1.11.2"
+version = "1.11.3.dev4"
```
javierdelapuente: This is new format for the library. Any reason for it?

requirements.txt
```diff
@@ -1,4 +1,5 @@
 cosl
+dpcharmlibs-interfaces
```
javierdelapuente: Can we pin this library?

examples/flask/test_rock/app.py
```diff
@@ -435,6 +436,29 @@ def redis_status():
     return "FAIL", 500
 
 
+def get_valkey_database() -> valkey.Valkey | None:
+    """Get valkey connection."""
+    if "valkey_db" not in g:
+        if "VALKEY_DB_CONNECT_STRING" in os.environ:
+            uri = os.environ["VALKEY_DB_CONNECT_STRING"]
+            g.valkey_db = valkey.Valkey.from_url(f"valkey://{uri}")
```
seb4stien: Is there a reason to  add "valkey://" here and not having it directly in VALKEY_DB_CONNECT_STRING? If feel the scheme is part of the URI we should provide to the apps.

Thanhphan1147: makes sense, valkey doesn't actually add that in the relation data, I'll update it

Thanhphan1147: I updated the general parsing logic to add "{prefix}://" This check if the scheme is there and add it if it's not there

examples/django/charm/requirements.txt
```diff
@@ -1,4 +1,5 @@
 cosl
+dpcharmlibs-interfaces
 jsonschema >=4.19,<4.20
 ops >= 2.6
-pydantic==2.6.4
+pydantic>=2.11,<3
```
javierdelapuente: Any reason not to pin to a specific version?

src/paas_charm/__init__.py
```diff
@@ -5,6 +5,8 @@
 
 """Module __init__."""
 
+import dpcharmlibs.interfaces  # noqa: F401
```
javierdelapuente: Is this necessary?

src/paas_charm/app.py
```diff
@@ -166,6 +167,32 @@ def generate_redis_env(relation_data: "PaaSRedisRelationData | None" = None) ->
...
+
+    Returns:
+        Valkey environment mappings if Valkey relation data is available, empty
+        dictionary otherwise.
+    """
+    if not relation_data:
+        return {}
+    prefix = "VALKEY"
+    return {
+        **_db_url_to_env_variables(prefix, str(relation_data.endpoints)),
```
javierdelapuente: Out of curiosity, does the endpoint include the username and password?

src/paas_charm/app.py
```diff
@@ -603,14 +633,20 @@ def _db_url_to_env_variables(prefix: str, url: str) -> dict[str, str]:
...
       database name extracted from the path
     """
-    prefix = prefix + "_DB"
-    envvars = _url_env_vars(prefix, url)
+    db_env_prefix = prefix + "_DB"
+    # Ensure the scheme in the url so that urllib can properly parse it into components.
+    # Providers like valkey sends the url without the scheme ( valkey_primary:6123 )
+    # This normalizes that URL to become valkey://valkey_primary:6123
+    if "://" not in url:
+        url = f"{prefix.lower()}://{url}"
```
javierdelapuente: In this case, the prefix and the scheme match. But it does not have to be true, for example, if it was using tls, it would be "valkeys"

(see https://valkey-py.readthedocs.io/en/latest/connections.html#valkey.Valkey.from_url)

src/paas_charm/valkey.py
```diff
@@ -0,0 +1,82 @@
...
+        automatically via model validators when the repository context is provided.
+
+        Raises:
+            InvalidValkeyRelationDataError: If invalid Valkey connection parameters were provided.
+
+        Returns:
+            ValkeyResponseModel with resolved secrets, or None if not ready.
+        """
+        try:
+            if len(self.valkey_interface.relations) == 1:
```
javierdelapuente: should be fail if there is more than one?

tests/unit/general/test_valkey.py
```diff
@@ -0,0 +1,158 @@
...
+                endpoints="valkey-host:6380",
+                username=None,
+                password=None,
+                tls=True,
+                tls_ca="-----BEGIN CERTIFICATE-----\nMIIBxTC...",
+                mode=None,
+                version="8.0.0",
+            ),
+            {
+                "VALKEY_DB_CONNECT_STRING": "valkey://valkey-host:6380",
```
javierdelapuente, line 58: I think this is wrong, if we use tls, the scheme should be valkeys.

src/paas_charm/valkey.py
alithethird: We need to block if we do not plan to support multiple relations.

tests/unit/requirements.txt
```diff
@@ -3,6 +3,6 @@ coverage[toml]
 gunicorn
 opentelemetry-api
 opentelemetry-sdk
-ops[testing]==2.21.1
+ops[testing]>=3.0
```
alithethird: We should pin this.

pyproject.toml
```diff
@@ -63,6 +63,7 @@ extend-exclude = '''
   |   examples/django/charm/lib/.*
   |   examples/fastapi/charm/lib/.*
   |   examples/go/charm/lib/.*
+  |   examples/go-valkey/charm/lib/.*
```
alithethird: `go-valkey` example doesn't exist

examples/django/charm/requirements.txt
Thanhphan1147: No specific reason besides that dpcharmlibs-interfaces pins it like this, do you think we should pin it exactly or follow the lib?

src/paas_charm/app.py
Thanhphan1147: Not for valkey, the username and password are passed separately ( we have the valkey_username and valkey_password attribute in the relation data iirc )

src/paas_charm/app.py
Thanhphan1147: Can we assume that the prefix will be `valkey://` for now since we say that we don't allow TLS yet?

tests/unit/general/test_valkey.py
Thanhphan1147: I'll remove this, the relation data parsing will block the charm if valkey is in TLS mode

Thanhphan1147: removed

src/paas_charm/__init__.py
Thanhphan1147: I assumed that we are doing import checks here so I added the lib by default, but now that I think about it since it's from pypi I'll just remove it. I guess this can be useful if anyone using 12-factor wants to customize their charm to be able to import the lib from the charm.py module.

Thanhphan1147: I removed this

pyproject.toml
Thanhphan1147: I removed it, thanks!

tests/unit/requirements.txt
Thanhphan1147: I removed it, thanks!

src/paas_charm/valkey.py
Thanhphan1147: Indeed, I updated the code to block if we have more than 1 valkey relation

requirements.txt
Thanhphan1147: updated

pyproject.toml
Thanhphan1147: removed

examples/django/charm/requirements.txt
javierdelapuente: I mean, it is for the test, so not very important, but I would rather pin it so it it a bit less flaky.

src/paas_charm/app.py
javierdelapuente: For postgres and redis for example we put the username and password in the connect string. In this case, it will not be there and will be in other env vars. I think 

For 26.04 we could change what we do know (maybe it is a good point to discuss), but I would rather be consistent in < 26.04.

javierdelapuente (approved):

src/paas_charm/valkey.py
```diff
@@ -0,0 +1,112 @@
...
+        Returns:
+            ValkeyResponseModel with resolved secrets, or None if not ready.
+        """
+        relations = self.valkey_interface.relations
+        if len(relations) > 1:
+            raise ValkeyMultipleRelationsNotSupportedError(
+                "Multiple valkey relations are not supported"
+            )
+        try:
+            if len(relations) == 1:
```
seb4stien: It feels strange to have the "if" inside the try. Shouldn't we have a "if len(relations) == 0" before the try and return None in such situation?

src/paas_charm/valkey.py
```diff
@@ -0,0 +1,112 @@
...
+                model = self.valkey_interface.interface.build_model(
+                    relation.id, component=relation.app
+                )
+                if not model.requests:
+                    return None
+                response = model.requests[0]
+                if response.tls:
+                    raise ValkeyTLSNotSupportedError("TLS mode is not supported for Valkey")
+                return response
+            return None
```
seb4stien: Do these lines have to be in the try block? (consider try/except/else if needed)

examples/django/charm/requirements.txt
Thanhphan1147: updated it to a == pin

src/paas_charm/app.py
Thanhphan1147: updated it to embed the username and password to the url ( which requires some additional logic )

tests/unit/general/test_http_proxy.py
```diff
@@ -160,6 +160,7 @@ def test_blocked_status_when_proxy_unavailable(
 
     base_state = request.getfixturevalue(base_state)
     base_state["config"] = config
+    base_state["leader"] = True
```
seb4stien, line 163: Is it related to this PR? (or fixing an existing issue?)

seb4stien (approved):

src/paas_charm/valkey.py
Thanhphan1147: thanks, I updated this part

src/paas_charm/valkey.py
Thanhphan1147: updated!

tests/unit/general/test_http_proxy.py
Thanhphan1147: the lib requires us to bump to ops 3.x, and the new version requires us to be leader to read the app databag.

PR #287, opened 2026-06-09
Set `alert_rules_path` of `LogForwarder` and `LogProxyConsumer` when using the v1 version of the Loki library

weiiwang01:
#### What this PR does

Set `alert_rules_path` of `LogForwarder` and `LogProxyConsumer` when using the v1 version of the Loki library

#### Why we need it

The missing `alert_rules_path` input in `LogProxyConsumer` causes the `cos_custom` custom Loki alert rules to be ineffective. Thank you @danielvnguyen for finding the cause of this issue.

#### Checklist

- [ ] I used AI to assist with preparing this PR
- [ ] I added or updated tests as needed (unit and integration)
- [ ] **If integration test modules are used:** I updated the workflow configuration  


srbouffard (approved):

danielvnguyen (approved):

danielvnguyen:
Thanks for fixing it!

src/paas_charm/observability.py
danielvnguyen, line 70: @weiiwang01  I have a question here. `enable_pebble_log_forwarding` returns True when the version of Juju is over **3.4**, right? 

For me locally, my Juju version is 3.6, which was installed automatically by Concierge on Multipass. So I think that I fall under this case. Are custom Loki alert rules not supported in this scenario?

weiiwang01: Yes, you are right, `alert_rules_path` need to be added here as well, thanks for pointing it out, updated.

alithethird (approved):

PR #335, opened 2026-08-10
Chore/refactor

alithethird:
#### What this PR does

- Simplifies `paas_charm` charm-library imports and integration initialization.
- Consolidates workload event handling around `reconcile()`, including migration-aware events.
- Handles invalid charm configuration and relation data centrally by setting a blocked status instead of allowing hook failures to escape.
- Simplifies integration-test fixtures and pins the reusable `charm-ci` workflow and `opcli` dependency to `v0.0.1-alpha.10`.
- Adds the OpenFGA v1 interface library required by the minimal Flask example.

#### Why we need it

The previous event handlers duplicated reconciliation and validation behavior across integrations. Centralizing this flow reduces repeated code, avoids constructing charm state twice for a single event, and keeps invalid configuration or relation data failures consistent and recoverable. Pinning the integration tooling also makes CI behavior reproducible.

#### Test plan

- Full unit suite: 508 passed, 13 skipped.
- Coverage report: 92%.
- Targeted lint, formatting, and type checks for the reconciliation changes passed.
- GitHub unit, documentation, and integration workflows are running for the latest branch revision.

#### Review focus

- Event-to-reconciliation wiring, especially which relation events rerun database migrations.
- Central handling of `CharmConfigInvalidError` and `RelationDataError` in `PaasCharm.reconcile()`.
- Removal of optional import fallbacks: required charm libraries are now imported eagerly.
- The pinned `charm-ci`/`opcli` version and the vendored OpenFGA interface library.

#### Potential breaking changes

Charm packages must include all directly imported charm libraries. Missing libraries now fail at import time rather than producing the previous compatibility warning. No public Python API is intentionally changed.

#### Dependencies and workflow changes

- Pins `canonical/charm-ci/.github/workflows/integration-test.yml` to `v0.0.1-alpha.10`.
- Pins `opcli` to `v0.0.1-alpha.10` in project dependencies and `uv.lock`.
- Adds `charms.openfga_k8s.v1.openfga` to the minimal Flask example.

#### Checklist

- [ ] I added or updated the documentation (not applicable: internal refactor)
- [ ] I updated `docs/changelog.md` (not applicable: no user-facing feature)
- [x] **If I added or changed integration tests:** I updated the charm-ci configuration as needed
- [ ] **If this is a Grafana dashboard:** I added a screenshot (not applicable)


src/paas_charm/charm.py
```diff
@@ -793,149 +735,21 @@ def _base_url(self) -> str:
...
-    @block_if_invalid_data
-    def _on_oauth_info_removed(self, _: ops.HookEvent) -> None:
-        """Handle the OAuth info removed event."""
-        self.restart()
-
-    @block_if_invalid_data
-    def _on_http_proxy_changed(self, _: ops.HookEvent) -> None:
-        """Handle http-proxy relation changed."""
-        self.restart()
+    def _reconcile_without_migrations(self, _: ops.RelationBrokenEvent) -> None:
```
javierdelapuente, line 753: Do we need this one? Why not just call reconcile directly?

src/paas_charm/charm.py
```diff
@@ -793,149 +735,21 @@ def _base_url(self) -> str:
...
-    @block_if_invalid_data
-    def _on_mysql_database_endpoints_changed(self, _: DatabaseRequiresEvent) -> None:
-        """Handle mysql's endpoints-changed event."""
-        self.restart(rerun_migrations=True)
-
-    @block_if_invalid_data
-    def _on_mysql_database_relation_broken(self, _: ops.RelationBrokenEvent) -> None:
-        """Handle mysql's relation-broken event."""
-        self.restart()
+    def _reconcile_with_migrations(self, _: DatabaseRequiresEvent) -> None:
```
javierdelapuente, line 749: I am not sure, but does this work with a lambda instead of creating this function?

javierdelapuente (approved):

src/paas_charm/charm.py
alithethird: `TypeError: Framework.observe requires a method as the 'observer' parameter, got <lambda>`. 
Ops requires observer callbacks to be bound methods and persists their method names for event replay.

swetha1654 (commented):
could you update the PR description please?

src/paas_charm/__init__.py
```diff
@@ -5,56 +5,10 @@
...
-        "Missing charm library, please run "
-        "`charmcraft fetch-lib charms.data_platform_libs.v0.data_interfaces`"
-    ) from import_error
-try:
-    import charms.redis_k8s.v0.redis  # noqa: F401
-except ImportError as import_error:
-    raise exceptions.MissingCharmLibraryError(
-        "Missing charm library, please run `charmcraft fetch-lib charms.redis_k8s.v0.redis`"
-    ) from import_error
+import charms.data_platform_libs.v0.data_interfaces  # noqa: F401
```
yanksyoon, line 8: Just out of curiosity to help me learn, are we raising import errors directly (it seems like the UX might be less friendly in terms of direct action)? Perhaps a rationale for # noqa above as a comment might be helpful

alithethird: yes we are.

But user will never get this because in the charmcraft extension we are injecting these libraries so they are always there.

We did not do it in v1 because users migth be using old versions of the extension and new version of paas-charm so the paas-charm expected library might not be injected by extension.

PR #345, opened 2026-08-19
feat: move application secret key out of peer relation

alithethird:
#### What this PR does

- Renames the `secret-storage` peer relation and interface to `peers`.
- Stores the generated application secret key in an application-owned Juju secret using the stable `app-secret-key` label across all frameworks.
- Restricts peer relation handling to peer coordination such as `PEER_FQDNS`.
- Routes leader election through holistic reconciliation: configuration and relation readiness are validated, the application secret is initialized when needed, and the workload is restarted without explicitly rerunning migrations.
- Defers the FastAPI Prometheus HTTP server to application lifespan, preventing Alembic imports from binding the metrics socket while retaining the separately configurable metrics port (`9464` in the example).
- Updates charm state, framework fixtures, Scenario tests, and documentation for the new storage model.

#### Why we need it

Application secret material should use Juju's secret storage rather than relation databags. Separating secret-key storage from peer coordination makes each helper's responsibility explicit and avoids coupling workload readiness to peer databag contents. Deferring the FastAPI metrics listener also prevents migration-time port collisions without coupling metrics to the application port.

#### Breaking changes

On upgrade, a fresh application secret key is generated because existing values are not migrated from the former peer databag. Applications that use this key for session signing may invalidate existing sessions. Charm metadata consumers must also use the renamed `peers` endpoint/interface.

#### Test plan

- `537 passed, 13 skipped` in the complete unit suite after synchronizing with current `main`
- Coverage: `93%`

#### Review focus

- Stable `app-secret-key` application-owned secret creation and rotation
- Holistic leader-election reconciliation without explicitly rerunning migrations
- Peer relation rename and coordination-only behavior
- FastAPI metrics startup lifecycle and separate configurable metrics port
- Upgrade impact described in `docs/changelog.md`

#### Dependencies and APIs

No new external dependencies. This introduces internal `SecretKeyStorage` and `Peers` helpers and removes the former peer-databag `SecretStorage` implementation.

#### Checklist

- [x] **If I added or changed integration tests:** I updated the charm-ci configuration as needed (`spread.yaml`, `integration-suites`, and/or `artifacts.yaml`)
- [x] **If this is a Grafana dashboard:** N/A


docs/reference/juju-events.rst
```diff
@@ -15,16 +15,19 @@ For a web app charm, the following events are observed:
...
+  * - ``config_changed``
     - Usually fired in response to a configuration change using the CLI.
     - Validate the charm configuration, run pending migrations and restart the workload.
-  * - ``secret_storage_relation_created``
-    - Fired when the relation is first created.
-    - Generate a new secret and store it in the relation data.
-  * - ``secret_storage_relation_changed``
+  * - ``leader_elected``
+    - Fired when a new leader unit is elected.
+    - Create the application secret key if it does not exist, then restart the workload.
```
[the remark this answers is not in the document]
alithethird: Implemented in `e8cf66d9`. Leader election now initializes a missing application secret without reconciling, which preserves the initial-install safeguard against migrations before relation data is complete. When the secret already exists and the workload service is running, it uses the normal `_reconcile()` path to validate readiness and restart the established workload. I also updated the event documentation and added coverage for new-secret initialization, a pre-start existing-secret state, and an established running workload.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

examples/fastapi/fastapi_app/metrics.py
```diff
@@ -0,0 +1,25 @@
...
+    server, thread = start_http_server(  # nosec
+        port=int(os.getenv("METRICS_PORT", "9464")),
+        addr="0.0.0.0",
+    )
+    try:
+        yield
+    finally:
+        server.shutdown()
+        thread.join()
+        server.server_close()
```
alithethird, line 25: Other wise the metrics server was running when the alembic migrations were running and failing the migration because the port is already in use.

src/paas_charm/charm.py
```diff
@@ -31,10 +31,11 @@
     LoggingFormat,
     read_paas_config,
 )
+from paas_charm.peers import Peers
 from paas_charm.rabbitmq import RabbitMQRequires
 from paas_charm.s3 import PaaSS3Requirer
 from paas_charm.saml import PaaSSAMLRequirer
-from paas_charm.secret_storage import KeySecretStorage
+from paas_charm.secret_key import SecretKeyStorage
```
alithethird, line 38: I renamed because it is not a storage for secrets anymore. It only holds the secret key.

examples/fastapi/fastapi_app/metrics.py
```diff
@@ -0,0 +1,25 @@
...
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
+
+
+@asynccontextmanager
+async def metrics_lifespan(_: object) -> AsyncIterator[None]:
+    """Run the Prometheus HTTP server only while the FastAPI application is running."""
+    from prometheus_client import start_http_server
+
+    server, thread = start_http_server(  # nosec
```
swetha1654: why the # nosec here?

examples/fastapi/fastapi_app/app.py
```diff
@@ -49,15 +49,14 @@
...
 root_path = f"/{parsed_url.path.strip('/')}" if parsed_url.path else ""
-app = FastAPI(root_path=root_path)
+app = FastAPI(root_path=root_path, lifespan=metrics_lifespan)
 app.add_middleware(SessionMiddleware, secret_key=os.environ.get("APP_SECRET_KEY"))
 app.add_middleware(ProxyHeadersMiddleware, trusted_hosts="*")
 
 set_tracer_provider(TracerProvider())
 get_tracer_provider().add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
 
 metrics.set_meter_provider(MeterProvider(metric_readers=[PrometheusMetricReader()]))
```
swetha1654: nit: a bit unrelated, but can we consider renaming this now that we have a dedicated metrics module? perhaps otel_metrics? for better readability.

src/paas_charm/charm.py
```diff
@@ -89,7 +90,8 @@ def __init__(self, framework: ops.Framework, framework_name: str) -> None:
         self._framework_name = framework_name
         self._paas_config = read_paas_config()
 
-        self._secret_storage = KeySecretStorage(charm=self, key=f"{framework_name}_secret_key")
+        self._secret_key = SecretKeyStorage(charm=self, label="app-secret-key")
```
swetha1654, line 93: would there be a reason to move from `{framework_name}_secret_key` to `app-secret-key`. I'm wondering if this would be a problem if there are 2 paas-charm applications in the same model with the same juju secret label?

tests/unit/general/test_oauth.py
```diff
@@ -302,7 +302,7 @@ def test_oauth_config_wrong_relation_order(
         remote_app_data={**OAUTH_RELATION_DATA_EXAMPLE, "client_secret_id": secret_id},
     )
     base_state["relations"].append(oauth_relation)
-    base_state["secrets"] = [testing.Secret(id=secret_id, tracked_content={"secret": "abc"})]
+    base_state["secrets"].append(testing.Secret(id=secret_id, tracked_content={"secret": "abc"}))
```
swetha1654, line 305: why are we appending now as opposed to initializing previously?

swetha1654 (approved):

src/paas_charm/charm.py
alithethird: Juju secret labels are charm-local, not model-global. Two applications in the same model can each own a different secret labeled `app-secret-key`; each secret has its own globally unique secret ID.

Renamed it to `app-secret-key` to keep the differences between frameworks to minimal.

tests/unit/general/test_oauth.py
alithethird: Previously we were storing the app-secret in the peer-relation. Now we let juju handle it.

In this line we are adding the secret for oauth. If we use `=` again, we are overriding the whole secrets the app has which was previously empty but not includes the `app-secret-key`

examples/fastapi/fastapi_app/metrics.py
alithethird: yeah this is not needed. thx!

PR #347, opened 2026-08-19
test: replace Redis examples with Valkey

alithethird:
#### What this PR does

Replaces Redis with Valkey across the integration test fixtures and example applications. Flask worker statistics and Celery connectivity now use the Valkey relation, while the Spring Boot example uses Spring Data Valkey with application-local Lettuce selection. Generic Spring Boot relation mappings remain client-agnostic so consumers can choose Valkey GLIDE, Lettuce, or Jedis. Obsolete Redis charm libraries, relations, fixtures, and tests are removed.

#### Why we need it

Redis support has been removed from the library, so the examples and integration coverage must use the supported `valkey_client` interface instead of referencing the retired Redis integration.

#### Checklist

- [x] **If I added or changed integration tests:** I updated the charm-ci configuration  
      as needed (`spread.yaml` `integration-suites` and/or `artifacts.yaml`)

#### Test plan

- Ran `tox -e fmt`.
- Collected the migrated Valkey and Flask worker integration tests with the integration tox environment.
- Parsed all example `charmcraft.yaml` files and compiled the changed Python files.
- Ran `.tox/unit/bin/pytest -q tests/unit/springboot/test_integrations.py tests/unit/springboot/test_charm.py` (14 passed).
- Ran `.tox/unit/bin/pytest -q tests/unit/flask/test_charm.py tests/unit/general/test_utils.py` (88 passed).
- Ran the Spring Boot example Maven tests with JDK 21 (3 passed).
- Ran `opcli artifacts build` successfully.
- Attempted `integration-test-local:ubuntu-24.04:build/tests/integration/run:integrations_test_valkey`; Spread backend preparation timed out before test execution because required rawfile CSI images could not be pulled from GHCR, leaving the Juju controller PVC pending.

#### Review focus

Please review the Flask Celery broker URL conversion: Celery still requires its Redis-compatible transport scheme while the workload receives credentials from `VALKEY_DB_CONNECT_STRING`. Spring Data also retains Redis-named Java APIs because they are the protocol-compatible client implementation used to connect to Valkey.

The generic Spring Boot mapping intentionally does not set `spring.data.valkey.client-type`. The example selects Lettuce in its own `application.properties` because it explicitly includes `lettuce-core`; other consumers remain free to choose another supported client.

This is a breaking change for example consumers that still relate through the `redis` endpoint or call `/redis/*`; they must move to the `valkey` relation and `/valkey/*` endpoints. No workflow changes are introduced.

#### SMTP behavior correction

The SMTP changes are unrelated to the Valkey refactor but correct the existing Spring Boot mappings. `spring.mail.properties.mail.smtp.auth` requires a boolean `true` or `false`, not authentication mechanism values such as `plain` or `none`. When credentials are absent, username and password are omitted so the workload does not receive `None@example.com` or null password values. The port is normalized to a string for Pebble environment values, and authenticated STARTTLS coverage was added. SMTP was bundled with the Valkey refactor, but this is a behavior correction rather than part of the Redis-to-Valkey migration.


seb4stien:
@alithethird shouldn't this one be "draft" for now?

src/paas_charm/springboot/charm.py
```diff
@@ -192,31 +193,32 @@ def generate_rabbitmq_env(
...
-    if parsed.username:
-        env["spring.data.redis.username"] = parsed.username
-    if parsed.password:
-        env["spring.data.redis.password"] = parsed.password
 
+    env = {
+        "spring.data.valkey.url": base_env["VALKEY_DB_CONNECT_STRING"],
+        "spring.data.valkey.host": base_env["VALKEY_DB_HOSTNAME"],
+        "spring.data.valkey.port": base_env["VALKEY_DB_PORT"],
+        "spring.data.valkey.client-type": "lettuce",
```
javierdelapuente: Do we need to set this?

src/paas_charm/springboot/charm.py
```diff
@@ -289,16 +291,21 @@ def generate_smtp_env(relation_data: "SmtpRelationData | None" = None) -> dict[s
...
-    return {
+    env = {
         "spring.mail.host": relation_data.host,
-        "spring.mail.port": relation_data.port,
-        "spring.mail.username": f"{relation_data.user}@{relation_data.domain}",
-        "spring.mail.password": relation_data.password,
-        "spring.mail.properties.mail.smtp.auth": relation_data.auth_type.value,
+        "spring.mail.port": str(relation_data.port),
+        "spring.mail.properties.mail.smtp.auth": str(
+            relation_data.auth_type.value != "none"
```
javierdelapuente, line 297: Was this a bug before? If so, I think we should describe it in the PR description (or create another PR).

javierdelapuente (commented):
A couple of small comments.

src/paas_charm/springboot/charm.py
alithethird: Yes this was a bug we did not realize before. It might be one of the reasons why smtp tests were kinda flaky:

- `smtp.auth` requires `"true"`/`"false"`, not `"plain"`/`"none"`.
- Missing credentials previously produced `"None@example.com"` and a `None` password.
- The port was normalized to a string for Pebble environment values.

src/paas_charm/springboot/charm.py
alithethird: We absolutely don't. I am removing it. AI add it there because it was using lettuce in the example charm, there are other alternatives to lettuce so we can not force people to use lettuce.

javierdelapuente (approved):

PR #348, opened 2026-08-21
test: modernize General, ExpressJS, and Django tests with Scenario Context

alithethird:
## Summary

- migrate the remaining General unit tests to Scenario `Context`/`State` fixtures
- improve ExpressJS Scenario context coverage for app environment and charm behavior
- improve Django Scenario context coverage for charm and worker behavior
- remove deprecated Harness, cwd-changing fixture, and package-wide `sys.path` mutation patterns from these three test scopes

## Validation

- `tox -e fmt`
- focused codespell, isort, Black, and pylint checks for all 19 touched Python files
- General + ExpressJS + Django unit tests, plus paas-config and Flask Gunicorn logging/webserver regressions, from the repository root: **439 passed, 13 skipped**
- the same combined regression set from `/tmp`: **439 passed, 13 skipped**


tests/unit/general/test_config.py
```diff
@@ -168,139 +173,161 @@
...
-        for substr in expected_log_message_substrs:
-            assert substr in str(exc).lower()
-        for substr in expected_status_message_substrs:
-            assert substr in str(blocked_harness.model.unit.status.message).lower()
-        for substr in unexpected_status_message_substrs:
-            assert substr not in str(blocked_harness.model.unit.status.message).lower()
+    context = context_factory(charm_type)
+    state_dict = framework_state_factory(charm_type)
+    if any("non-optional-string" in config for config in required_configs):
+        state_dict["config"].pop("non-optional-string", None)
```
swetha1654, line 204: what exactly are we doing here?

tests/unit/general/test_oauth.py
```diff
@@ -4,14 +4,12 @@
 """Charm unit tests for Oauth relation."""
 
 # Very similar cases to other frameworks. Disable duplicated checks.
-# pylint: disable=R0801
+# pylint: disable=R0801,C0302
```
swetha1654: could you use the rule name instead of id here?

tests/unit/scenario.py
swetha1654, line 1: why not move this whole file to conftest.py?

swetha1654 (approved):
LGTM, thanks!

tests/unit/general/test_config.py
alithethird: we have `non-optional-string` config option in the actual example charms but not in every one of them. To keep the tests concise we are removing it from the charm config options to make sure we are only testing the non optional config options injected by the test here.

tests/unit/general/test_oauth.py
alithethird: Updated the file-level disable to use the symbolic Pylint names `duplicate-code` and `too-many-lines`.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

tests/unit/scenario.py
alithethird: Moved the Scenario root mappings and resolver into `tests/unit/conftest.py` and deleted `tests/unit/scenario.py`. Tests now access the resolver through normal pytest fixture discovery, without importing conftest directly.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

PR #359, opened 2026-09-16
12-factor DEV API for Custom Relations

canon-cat:
#### What this PR does

This PR implements an API that lets a 12-factor charm author add their own Juju relations and have them participate in the framework lifecycle exactly like a built-in relation, without importing or subclassing internal classes. See ISD291.

#### Why we need it

The 12-factor tooling (paas-charm) does not currently support adding Juju relations beyond the set of relations built into the framework.

#### Checklist

- [ ] **If I added or changed integration tests:** I updated the charm-ci configuration  
      as needed (`spread.yaml` `integration-suites` and/or `artifacts.yaml`)


alithethird (approved):

docs/reference/custom-relations.rst
```diff
@@ -0,0 +1,134 @@
...
+targets exactly one framework, so the author always knows which environment
+variable names that framework's workload expects and can emit them directly.
+Per-framework remapping for custom relations is out of scope.
+
+Optional charm libraries
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+A custom relation's requirer typically imports an author-supplied charm
+library. Providing that library is the author's responsibility (for example
+via ``charmcraft fetch-lib``). The framework does **not** wrap ``setup()``,
```
javierdelapuente, line 131: `charmcraft fetch-lib` is deprecated, so not sure it is a good example.

src/paas_charm/charm.py
```diff
@@ -396,6 +404,56 @@ def _init_oauth(self, requires: dict[str, RelationMeta]) -> "PaaSOAuthRequirer |
...
+        if not self.custom_relations:
+            return []
+        context = self._build_custom_relation_context()
+        requires: dict[str, RelationMeta] = self.framework.meta.requires
+        relations: list[CustomRelation] = []
+        for relation_class in self.custom_relations:
+            if not isinstance(relation_class, type) or not issubclass(
+                relation_class, CustomRelation
+            ):
+                logger.warning(
```
javierdelapuente, line 441: Should we crash in here?

src/paas_charm/charm.py
```diff
@@ -396,6 +404,56 @@ def _init_oauth(self, requires: dict[str, RelationMeta]) -> "PaaSOAuthRequirer |
...
+                    "Skipping non-CustomRelation entry in custom_relations: %r",
+                    relation_class,
+                )
+                continue
+            relation_name = relation_class.relation_name
+            if relation_name not in requires:
+                continue
```
javierdelapuente, line 448: Should we crash in here? At least we should log an error.

src/paas_charm/charm.py
```diff
@@ -651,6 +716,9 @@ def _reconcile(self, rerun_migrations: bool = False) -> None:
                 self._oauth.update_client()
             self.update_app_and_unit_status(ops.MaintenanceStatus("Preparing service for restart"))
             self._create_app().restart()
+            for relation in self._custom_relations:
+                if relation.is_ready():
+                    relation.reconcile()
```
javierdelapuente, line 721: The reconcile method in the CustomRelation is not clear.

Going back to the `add-custom-relation.rst` it says:
```
* run its own side-effect reconcile logic (for example publishing routing
  configuration to a proxy, or pushing generated config files into the
  container).
```

However, the reconcile is called after the workload is restarted, so it is not clear what "pushing generated config files" means (they are sent when the workload can be working). 

I would remove the reconcile function, as it is unclear when it is run, and instead add hooks in the future if they are needed/requested, like "run_before_workload_starts" or "run_after_workload_starts".

tests/integration/flask/test_temporal.py
```diff
@@ -0,0 +1,116 @@
...
+
+TEMPORAL_APP = "temporal-k8s"
+TEMPORAL_ADMIN_APP = "temporal-admin-k8s"
+POSTGRESQL_APP = "postgresql-k8s"
+CERTIFICATES_APP = "self-signed-certificates"
+TEMPORAL_HOST = "temporal.local.test"
+TEMPORAL_PORT = "7233"
+
+
+def _deploy_temporal(juju: jubilant.Juju) -> None:
```
javierdelapuente, line 20: Following other tests, I would put this in a fixture.

tests/unit/test_custom_relations_api.py
```diff
@@ -0,0 +1,279 @@
...
+    return context_factory(OverwritingCharm)
+
+
+@pytest.fixture
+def side_effect_context(context_factory) -> testing.Context:
+    """Return a Context rooted at the side-effect custom relation charm."""
+    return context_factory(SideEffectCharm)
+
+
+# pylint: disable=redefined-outer-name
```
javierdelapuente, line 117: why this?

src/paas_charm/charm.py
```diff
@@ -396,6 +404,56 @@ def _init_oauth(self, requires: dict[str, RelationMeta]) -> "PaaSOAuthRequirer |
...
+
+        Returns:
+            A :class:`Context` populated with the current charm configuration.
+        """
+        return Context(
+            app_name=self.app.name,
+            framework_name=self._framework_name,
+            port=self._workload_config.port,
+            container_name=self._workload_config.container_name,
+            config=self._resolve_charm_config(),
```
javierdelapuente, line 418: _resolve_charm_config calls `get_content(refresh=True)`.

IINM, this is not a free call. Refreshing a secret means a call to the uniter and then to the Juju controller.

tests/unit/test_custom_relations_api.py
canon-cat: Without it lint complains about 

```
tests/unit/test_custom_relations_api.py:118:34: W0621: Redefining name 'envvar_context' from outer scope (line 88) (redefined-outer-name)
```

canon-cat: Will rename fixtures to avoid suppressing this error.

src/paas_charm/charm.py
canon-cat: _resolve_charm_config is a simple refactoring of repeated code in the existing [charm.py](https://github.com/canonical/paas-charm/blob/090bbf832f5afd870cd7e8c5389e4746bc3c1999/src/paas_charm/charm.py) (line numbers for 090bbf832f5afd870cd7e8c5389e4746bc3c1999):

[Lines 410 - 417](https://github.com/canonical/paas-charm/blob/090bbf832f5afd870cd7e8c5389e4746bc3c1999/src/paas_charm/charm.py#L410):
```
        charm_config = {k: config_get_with_secret(self, k) for k in self.config.keys()}
        config = typing.cast(
            dict,
            {
                k: v.get_content(refresh=True) if isinstance(v, ops.Secret) else v
                for k, v in charm_config.items()
            },
        )
```

[Lines 688 - 695](https://github.com/canonical/paas-charm/blob/090bbf832f5afd870cd7e8c5389e4746bc3c1999/src/paas_charm/charm.py#L688-L696):
```
        charm_config = {k: config_get_with_secret(self, k) for k in self.config.keys()}
        config = typing.cast(
            dict,
            {
                k: v.get_content(refresh=True) if isinstance(v, ops.Secret) else v
                for k, v in charm_config.items()
            },
        )
```

src/paas_charm/charm.py
alithethird: Good catch! This should be moved to before restarting the app.
We can explain in the docs and in docstring that it is run right before the restart and after migrations.

Would there be a need for running it before migrations? @javierdelapuente

```suggestion
            for relation in self._custom_relations:
                if relation.is_ready():
                    relation.reconcile()
            self.update_app_and_unit_status(ops.MaintenanceStatus("Preparing service for restart"))
            self._create_app().restart()
```

src/paas_charm/charm.py
alithethird: This should be an Error and we should raise here.

src/paas_charm/charm.py
alithethird: Error for custom relations, do nothing for built-ints

docs/reference/custom-relations.rst
canon-cat: Removed mentions of fetch-lib

src/paas_charm/charm.py
canon-cat: Non-CustomRelation class raises an error now.

src/paas_charm/charm.py
canon-cat: Per our conversation, `reconcile()` has been removed.

tests/integration/flask/test_temporal.py
canon-cat: Updated to use fixture.

tests/unit/test_custom_relations_api.py
canon-cat: Updated, `disable=redefined-outer-name`  is no longer necessary.

src/paas_charm/charm.py
canon-cat: We have added one more call, but the new call path does not involve loops. Updating the existing logic is, probably, out of scope for this PR.

src/paas_charm/charm.py
canon-cat: For now we raise an error if custom relation exists but not referenced.

PR #360, opened 2026-09-18
fix: respect PostgreSQL TLS relation data

alithethird:
## Summary

- read the PostgreSQL provider's `tls` relation field when generating connection URIs
- append `sslmode=disable` or `sslmode=require` while preserving an explicit existing `sslmode`
- add durable unit coverage for TLS enabled, TLS disabled, absent TLS data, explicit SSL modes, and non-PostgreSQL compatibility
- bump the package minor version from 1.12.3 to 1.13.0

This fixes the upstream cause identified in canonical/mattermost-k8s-operator#42.

## Disposable integration proof

The live end-to-end scenario is intentionally isolated in the stacked draft PR #362. That PR deploys Flask with PostgreSQL, verifies `sslmode=disable` and live connectivity, enables TLS through `self-signed-certificates` and verifies `sslmode=require` plus connectivity, then removes TLS and verifies the connection returns to `sslmode=disable` while remaining usable.

The proof PR exists only to demonstrate the behavior in a real Juju environment. It should remain draft and should not be merged, so this temporary scenario does not become a permanent maintenance burden in the v1 integration suite.

## Validation

- `tests/unit/flask/test_databases.py`: 16 passed
- full unit suite: 513 passed, 13 skipped
- changed Python files pass Black, isort, pylint, and pydocstyle
- live integration proof: #362
- `test_postgresql_connection_string_respects_tls` executed successfully in CI; `flask_test_database` completed with `2 passed in 413.66s` ([job](https://github.com/alithethird/paas-charm/actions/runs/35320267902/job/105521219573))

## Target

Base branch: `v1`


src/paas_charm/databases.py
```diff
@@ -39,17 +40,43 @@ class PaaSDatabaseRelationData(BaseModel):
...
+        """Add the PostgreSQL SSL mode indicated by the relation data.
+
+        Args:
+            uri: The PostgreSQL connection URI.
+            tls: Whether TLS is enabled on the database provider.
+
+        Returns:
+            The connection URI with the appropriate SSL mode.
+        """
+        if self.relation_name != "postgresql" or tls is None:
```
swetha1654, line 53: is there a reason we are honoring this field only for the postgresql relation, from what i understand in the library, it is applicable to all the database relations in the library. (for eg mysql)

alithethird: The shared `data_interfaces` library exposes generic helpers such as `set_tls`, which can make TLS appear applicable to every database relation. However:



- The current `postgresql_client` provider schema explicitly defines `tls`.

- The current `mysql_client` and `mongodb_client` provider schemas do **not** define `tls`.

- `sslmode=require|disable` is PostgreSQL/libpq-specific. MySQL and MongoDB use different connection parameters and semantics, so applying this transformation generically would produce incorrect URIs.

- The current MySQL and MongoDB operators do not publish `tls` through these client relations.

